PR #4, opened 2023-03-31
improve `coverage`

brunoerg:
1. Replace `split(" ")` to `split()` to avoid redundance

2. Use `DataFrame` constructor

3. Add (addr & net_masks) to ensure the bitwise AND operation is evaluated before the comparison with `network_addresses`

4. There is no need to use `len` with `.index`, their length can be got directly


fjahr:
ACK, thanks!

PR #6, opened 2023-11-21
Flakify

jurraca:
Blocked by [PR](https://github.com/fjahr/rpki-client-nix/pull/4) in RPKI-client repo.

Takes the exact devShell we had before, but adds a static version of rpki-client via our flake, and sets Python version to 310 and pins nixpkgs version. This should help prevent reproducibility issues due to tooling (python and python package versions). 

Workflow would be: clone repo, run `nix develop`, and then run commands as specified in README.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,166 @@
+{
+  "nodes": {
+    "nixpkgs": {
+      "locked": {
+        "lastModified": 1700403855,
+        "narHash": "sha256-Q0Uzjik9kUTN9pd/kp52XJi5kletBhy29ctBlAG+III=",
+        "owner": "NixOS",
+        "repo": "nixpkgs",
+        "rev": "0c5678df521e1407884205fe3ce3cf1d7df297db",
+        "type": "github"
+      },
+      "original": {
+        "owner": "NixOS",
+        "ref": "nixos-23.05",
+        "repo": "nixpkgs",
+        "type": "github"
+      }
+    },
+    "nixpkgs_2": {
+      "locked": {
+        "lastModified": 1700403855,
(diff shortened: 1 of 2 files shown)


jurraca:
@fjahr Added comments and docs. I'm not sure it's entirely appropriate yet to add this as recommended method without further testing from different target OSs. You tested on mac silicon (aarch) ? 
lmk what you think of the `nix.md` as background/reasoning. 
Finally, I'll need one last commit to update `flake.lock` to point to our nixified `rpki-client` repo when that's merged, so blocked until then.

fjahr:
@jurraca Looks great to me! But yeah, I think we can still word it a bit more softly. Something like "If you know nix or want to learn it can save you a lot of work." I will think will make a suggestion for a bit of rewording. But since I have learned now that outdated rpki-client versions are not available through the usual package managers, we may run into issues rather quickly if there are any breaking changes in a new version.

I am using apple silicon but I will test it all more thoroughly once the rpki-client-nix PR is merged :)

fjahr:
@jurraca Here is a bit of rewording plus some additional information in the readme, what do you think? https://github.com/fjahr/kartograf/commit/47c688edd7f14375ad1c546573d8e65c0de2d0df In terms of structure, I put the nix instructions on the level of the OSs, and not inside rpki-client, since using nix also takes care of the python dependencies. In case users want to not use nix, I added some additional information on versioning.

Will do some additional testing now. If you like the changes you call pull them in here or I can do a follow-up as well, either works for me.

jurraca:
nice, yep that works. I added your commit above [1cf235e](https://github.com/fjahr/kartograf/pull/6/commits/1cf235e0cb6d212186a684febba8dad927e1b6f1). 
Also removed the `shell.nix` file we don't need anymore.

jurraca:
uh i'm an idiot didn't see your commit land in this branch. let's squash this before we merge it.

PR #7, opened 2023-12-06
Add nix support for added pandarallel

fjahr:
I have added parallelization of the merge process using `pandarallel` with the last commit in master. I have also attempted to add this to our `flake.nix` but so far I haven't been successful.

This is the package I am using (v 1.6.5): https://pypi.org/project/pandarallel/

It appears there is no nix pkg for this. So I tried to add it's build to our `flake.nix` directly. It appears to work but when I try to use kartograf in the `nix develop` shell it appears to not be available.

Could you please take a look at the approach and code @jurraca ? Thanks a lot!

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -17,7 +17,30 @@
     # Build for all default systems: ["x86_64-linux" "aarch64-linux" "x86_64-darwin" "aarch64-darwin"]
     utils.lib.eachDefaultSystem (system: let
       pkgs = nixpkgs.legacyPackages.${system};
+
+      # Custom derivation for pandarallel
+      pandarallel = pkgs.python3Packages.buildPythonPackage rec {
+        pname = "pandarallel";
+        version = "1.6.5";
+
+        src = pkgs.python3Packages.fetchPypi {
+          inherit pname version;
+          sha256 = "HC35j/ZEHorhP/QozuuqfsQtcx9/lyxBzk/e8dOt9kA=";
+        };
+
+        propagatedBuildInputs = with pkgs.python3Packages; [ pandas dill psutil ];
+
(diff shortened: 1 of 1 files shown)


jurraca:
ok, so i fixed the expression a bit, and got an error about pandarallel being unable to find numpy, seemingly because the package expects python 3.11. 

However, this works if you just bump everything in the expression to python 3.11 . See [PR](https://github.com/fjahr/kartograf/pull/8) into this branch.
We _could_ override it to accept python 3.10 but I haven't attempted--could get complicated.
Do you have a requirement for 3.10 over 3.11 or is it the same to you? 

I should mention that packaging python libs for usage with Nix is really hard, mostly due to the way Python links libs. I stay away from it as much as possible :upside_down_face: . In this case we got lucky because all the libs before this were already packaged in nixpkgs. 

def working though, from `htop` while running `./run map` 
image

and here's the log output . 25 mins :tada: 
```
--- Fetching RPKI ---

Downloading RPKI Data
...finished in 0:01:48.848190

--- Validating RPKI ---

Validating RPKI ROAs
116503 raw RKPI DB entries validated and saved to /home/base/code/kartograf/out/1701880411/rpki/rpki_raw.json
...finished in 0:21:12.303198

--- Parsing RPKI ---

Parsing 116503 ROAs
Output: 361231
Duplicates: 34089
Invalids: 20284
Incompletes: 0
...finished in 0:01:15.929132

--- Sorting results ---

...finished in 0:00:03.099646

--- Finishing Kartograf ---

The SHA-256 hash of the result file is: c22f057a4412fe5467149806316652332bba7ada0bbdc3772f83a02e3087bac1
Total runtime: 0:24:20.193839
```

PR #8, opened 2023-12-06
fix expression, bump to 3.11 to get pandarallel to work

jurraca:
Pandarallel wants python 3.11. If we just build the package and pass it to our environment the python interpreter can find it and run it. 
The `python31x.withPackages` construct is for packages that are already in Nixpkgs, so when it loads them it makes sure that it has the right version for all the packages and that they'll work. 
In this case, we just declare `pandarallel` with python 311 independently and pass it in. 

We could find a working version of pandarallel with 310 if needed but it would take more work. 
Bumping everything to 311 works though.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -23,7 +23,7 @@
         pname = "pandarallel";
         version = "1.6.5";
 
-        src = pkgs.python3Packages.fetchPypi {
+        src = pkgs.python311Packages.fetchPypi {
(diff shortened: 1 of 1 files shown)


fjahr:
Looks good, I will finish with merging everything until I have finished a full reproduction but the error on startup is definitely fixed. Thanks a lot!

PR #9, opened 2023-12-08
Add coordinated launch feature (-wait option)

fjahr:
This allows multiple contributors to launch the mapping process simultaneously which would hopefully result in mostly the same data for all of them, potentially even let some of them get the exact same result, improving confidence in the correctness of the data collected.

It is currently unclear how well this works, so it needs to be tested extensively.

Usage example:
```
$ ./run map -w=1702047310

--- Start Kartograf ---

Using rpki-client version 8.6.
Coordinated launch mode: Waiting until 1702047310 (2023-12-08 15:55:10 CET) to launch mapping process.
[waits...]
The epoch for this run is: 1702047310 (2023-12-08 14:55:10 UTC, local: 2023-12-08 15:55:10 CET)
```

### DIFF
--- a/kartograf/collectors/parse.py
+++ b/kartograf/collectors/parse.py
@@ -4,9 +4,10 @@ from kartograf.timed import timed
 
 @timed
 def parse_routeviews_pfx2as(context):
-    raw_file = f'{context.data_dir_collectors}pfx2asn.txt'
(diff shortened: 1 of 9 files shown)


fjahr:
I have done several tests of this feature over the last few days and the results have looked very promising. I will merge now since I think it will certainly be valuable, though we still need a bit more testing to decide if it will be reliable enough to be part of major workflows.

PR #10, opened 2023-12-15
fetch: expose `rpki-client` logs in `fetch_rpki_db`

brunoerg:


fjahr:
Thanks, could you hide this behind a debug option? The logs from `rpki-client` are very noisy but that isn't an issue usually, so I would not like to print them by default.

fjahr:
FYI: Here you can see the logs of the periodical runs of `rpki-client` from one of the maintainers: https://console.rpki-client.org/ so this is kind of the baseline that appears not be worrying anyone ;)

He also writes: `Note: rpki-client outputs information about errorenous objects and problems with repositories.
Any errors in the below log should be solved by CA operators!`

brunoerg:
> FYI: Here you can see the logs of the periodical runs of rpki-client from one of the maintainers: https://console.rpki-client.org/ so this is kind of the baseline that appears not be worrying anyone ;)

> He also writes: Note: rpki-client outputs information about errorenous objects and problems with repositories. Any errors in the below log should be solved by CA operators!


Hmm, fair enough. So these logs are not so helpful as I thought. Gonna close it.

fjahr:
@brunoerg I still think it may be interesting to add a debug parameter that logs a lot more details, but it's probably not highest priority. But once we have it we should include logging this as well when it's set.

PR #13, opened 2024-07-03
Add NixOS module to facilitate scheduled runs

virtu:
Adds a NixOS module to facilitate scheduled runs. Removing any user action should increase the number of participants in community runs. 

Deployment via NixOS is as simple as:

```nix
services.kartograf = {
    enable = true;
    workers = 2;
};
```

The default schedule is monthly (`*-*-01 00:00:00 UTC`), but this is easily changed (e.g., `services.kartograf.schedule = weekly;`). By default,  IRR and RV data is used; this is can be changed via the `useIRR` and `useRV` settings.

I tried to avoid changes to kartograf code as much as possible, but some were necessary:
- Introduce a `-s`/`--silent` option, which is enabled in the module by default, to suppress output from `pandarallel` progress bar to avoid spamming the systemd journal with multiple BLOB entries per second.
- Introduce a `-p`/`--workers` option to set the number of threads for `pandarallel` to avoid hogging the CPU. (Since `pandarallel` is only used in the `merge` part, I don't see a problem with reproducibility when slowing down this part.)

The systemd job takes care of removing all data produced by kartograf except the final result to avoid filling up the file system (otherwise each run would contribute multiple GB of data, even when the `--clean` option is used). To avoid further changes to kartograf's code minimal, this task is handled via a systemd post-execution script.

P.S. I did not apply `nixpkgs-fmt` to the existing `flake.nix` file because I was not sure if the style was produced with a nix pretty printer I do not know. If it wasn't, let me know if it's ok to format the file.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -14,6 +14,8 @@
     utils,
     rpki-cli,
   }:
+    # Add a kartograf module for NixOS (see module.nix for details)
+    { nixosModules.kartograf = import ./module.nix self; } //
     # Build for all default systems: ["x86_64-linux" "aarch64-linux" "x86_64-darwin" "aarch64-darwin"]
     utils.lib.eachDefaultSystem (system: let
       pkgs = nixpkgs.legacyPackages.${system};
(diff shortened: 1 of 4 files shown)


flake.nix
```diff
@@ -38,21 +40,65 @@
...
+          buildInputs = with pkgs.python3Packages; [
             rpki-cli.defaultPackage.${system}
-            # Python 3.10 with packages
-            (pkgs.python311.withPackages (ps: [
-              ps.pandas
-              ps.beautifulsoup4
-              ps.requests
-              ps.tqdm
-            ]))
+            pkgs.python3
```
jurraca, line 70: nice, i guess the deps gets linked to the right python version without setting it explicitly like at line 53 ? If so could you just use the same package declaration in both places?

flake.nix
```diff
@@ -14,6 +14,8 @@
     utils,
     rpki-cli,
   }:
+    # Add a kartograf module for NixOS (see module.nix for details)
+    { nixosModules.kartograf = import ./module.nix self; } //
```
jurraca, line 18: fine to throw this in the body of the outputs, if you want

module.nix
```diff
@@ -0,0 +1,85 @@
...
+    mv out/$timestamp/final_result.txt ${cfg.resultPath}/asmap-$timestamp.txt
+    echo "Copied result from /out/$timestamp/final_result.txt to ${cfg.resultPath}/asmap-$timestamp.txt"
+    rm -rf data out
+    echo "Cleaned up temporary directories."
+  '';
+in
+{
+  options.services.kartograf = {
+    enable = mkEnableOption "kartograf";
+    clean = mkEnableOption "cleaning up of temporarty artifacts after processing." // { default = true; };
```
jurraca, line 20: sp "temporary"

jurraca (approved):
nice work, thanks for doing this. A few comments.

flake.nix
virtu: I applied this kludgy merge approach (which I also use in my own flakes) because I never found a way to get this working with flake-utils `lib.eachDefaultSystem`.

If I put the module in the output body, I end up with

```
git+file:///home/virtu/code/kartograf
├───devShell
│   ├───aarch64-darwin: development environment 'nix-shell'
│   ├───aarch64-linux omitted (use '--all-systems' to show)
│   ├───x86_64-darwin omitted (use '--all-systems' to show)
│   └───x86_64-linux omitted (use '--all-systems' to show)
├───nixosModules
│   ├───aarch64-darwin: NixOS module
│   ├───aarch64-linux: NixOS module
│   ├───x86_64-darwin: NixOS module
│   └───x86_64-linux: NixOS module
└───packages
    ├───aarch64-darwin
    │   ├───default: package 'kartograf-1.0.0'
    │   └───kartograf: package 'kartograf-1.0.0'
    ├───aarch64-linux
    │   ├───default omitted (use '--all-systems' to show)
    │   └───kartograf omitted (use '--all-systems' to show)
    ├───x86_64-darwin
    │   ├───default omitted (use '--all-systems' to show)
    │   └───kartograf omitted (use '--all-systems' to show)
    └───x86_64-linux
        ├───default omitted (use '--all-systems' to show)
        └───kartograf omitted (use '--all-systems' to show)
```

instead of 

```
git+file:///home/virtu/code/kartograf
├───devShell
│   ├───aarch64-darwin: development environment 'nix-shell'
│   ├───aarch64-linux omitted (use '--all-systems' to show)
│   ├───x86_64-darwin omitted (use '--all-systems' to show)
│   └───x86_64-linux omitted (use '--all-systems' to show)
├───nixosModules
│   └───kartograf: NixOS module
└───packages
    ├───aarch64-darwin
    │   ├───default: package 'kartograf-1.0.0'
    │   └───kartograf: package 'kartograf-1.0.0'
    ├───aarch64-linux
    │   ├───default omitted (use '--all-systems' to show)
    │   └───kartograf omitted (use '--all-systems' to show)
    ├───x86_64-darwin
    │   ├───default omitted (use '--all-systems' to show)
    │   └───kartograf omitted (use '--all-systems' to show)
    └───x86_64-linux
        ├───default omitted (use '--all-systems' to show)
        └───kartograf omitted (use '--all-systems' to show)
```

If you know a workaround or were thinking of another place to put it, let me know.

flake.nix
jurraca: yep, nice.

flake.nix
jurraca: ah I see, yea that's a limitation of flake-utils. the merge approach is fine.

jurraca:
ACK

fjahr:
ACK, thanks @virtu !

PR #17, opened 2024-08-19
Update flake: bump rpki-client and flake-utils versions

jurraca:
Bumps the `flake.lock` to use the latest rpki-client versions. Also bumps `flake-utils` and the `nixpkgs` rev (but not version).


fjahr:
ACK, verified the `rpki-client-nix` hash.

PR #21, opened 2024-09-23
Add debug logging and log rpki-client output to it

fjahr:
We sometimes see issues with `rpki-client` which lead to mismatches in our results. Logging the output from `rpki-client` to a debug.log file should make it easier for us to debug these issues and get help from upstream.

### DIFF
--- a/kartograf/context.py
+++ b/kartograf/context.py
@@ -83,3 +83,5 @@ class Context:
         self.final_result_file = f"{self.out_dir}final_result.txt"
 
         self.max_encode = self.args.max_encode
+
+        self.debug_log = f"{self.out_dir}debug.log"
--- a/kartograf/context.py
+++ b/kartograf/context.py
@@ -84,4 +84,7 @@ class Context:
 
         self.max_encode = self.args.max_encode
 
-        self.debug_log = f"{self.out_dir}debug.log"
+        if self.args.debug:
+            self.debug_log = f"{self.out_dir}debug.log"
+        else:
+            self.debug_log = ""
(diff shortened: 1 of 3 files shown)


fjahr:
@jurraca could you take a look when you have a moment?

brunoerg:
Concept ACK

fjahr:
> Might make sense to have it behind a flag though, what do you think? I can see it being useful to collect everybody's debug logs for every generation session we run, but i'd generally like to limit the amount of data we write to every user's machine since we already generate a ton. so maybe this is good for now and we think about a flag later.

Generally in the future it makes sense to put it behind a flag, especially when we add more stuff that is less essential. I have added a debug flag but for now the users can not unset it. The reason is that we really want this debugging information to investigate if we have issues `rpki-client` again. That seems to be the most fragile part of our setup currently and I don't want the information to be lost because someone forgot the flag. Almost everything else we can recover with our reproducibility but what happens in `rpki-client`, especially networking issues, we need to capture during the run. When we are more stable we can make it optional. And for now the amount of data written is 1-2MB, that isn't that much in the grand scheme of things.

jurraca:
ACK. 

More an fyi, this seems to produce a fair amount of duplicate logs (3874 unique lines vs 12459 total in my latest run).

fjahr:
> More an fyi, this seems to produce a fair amount of duplicate logs (3874 unique lines vs 12459 total in my latest run).

You mean output from rpki-client, right? Maybe would be something to report/ask about upstream?

PR #22, opened 2024-09-26
nix: Fix links after move to asmap org

fjahr:
While the redirects work there is no guarantee they will work forever so we should fix the links.


jurraca:
ACK. Tested and builds as expected.

Issue #19, opened 2024-08-21
rpki-client: Inconsistent RPKI repository caching results


fjahr:
We are sometimes getting more RPKI repositories than on other runs. IRR always completes the data but it would be good to understand what is causing this and if there is anything that we can improve on our side.

I opened an issue in the rpki-client github: https://github.com/rpki-client/rpki-client-portable/issues/116

job:
It is best to use the same repository across multiple runs, not start empty every time

job:
Another thing that stands out: the `-P` option is used, this option is only intended for very special use-cases; and the wrong value could result in repositories going missing. Why is `-P` used?

fjahr:
Ok, I think we can try to add something like this for our scheduled runs. We can try to sync once before the we actually run and then just resync with the same cache dir when our run actually starts.

job:
I recommend to somehow sync at least once an hour, this increases the chance to overcome intermittent transient reachability issues repositories might have

jurraca:
iiuc, is this actually an issue? If we want a static set of RPKI repositories, we could use the `-H` flag, which takes a list of domains to use for fetching RPKI data. That way we would have the same repositories every time. But I don't necessarily see the issue with letting `rpki-client` fetch RPKI data for us as it sees fit.

fjahr:
Agree, in the latest runs we have had pretty much no mismatches so it doesn't seem to be an actual problem (anymore). Good to know we have that option with the list but given the recent results I would say we don't do that unless we have to since the list itself would be another thing we have to manage.

PR #25, opened 2024-09-29
Warm up rpki-client cache 30 min before scheduled run

fjahr:
This is an idea to improve the issue described in #19. We warm up the rpki-client cache once 30min before the scheduled run.

Keeping this as a draft for now. I need to think about if we want to do some more refactorings, if this is really helping and if it is stable. I don't think we need to have this soon. It would be good enough for the next runs in the short-term that we have debug logging and see if the issue #19 comes up more regularly. If it comes up more regularly maybe we see some other issues in those logs that give us a better idea to actually fix the root cause and this might turn out to not be needed at all.

### DIFF
--- a/kartograf/kartograf.py
+++ b/kartograf/kartograf.py
@@ -1,3 +1,4 @@
+import asyncio
 import datetime
 from datetime import timezone
 import shutil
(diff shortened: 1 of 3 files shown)


jurraca:
Would it make sense to fetch the repository data regularly and maintain the latest set in the asmap-data repo?

fjahr:
> Would it make sense to fetch the repository data regularly and maintain the latest set in the asmap-data repo?

No, I don't think that makes sense. First of all, we would need to pay for the storage space we would need for that. And I also don't see any upside to it. The whole RPKI repository infrastructure is optimized for (decentralized) data availability so if there is a reason why we can't get the data from it reliably we should first find out what the issue is. GitHub is also not a very reliable service IMO so we would need to think about redundancy etc. and we would also make RPKI a centralized data source because if something goes wrong only our data might be used. I don't want to risk that.

fjahr:
Closing in favor of #78

PR #31, opened 2024-10-30
add diff command

jurraca:
takes two ASmaps and compares them (new, removed, and changes/transfers). It prints the counts of each, and writes the full set to an output file.

This is similar to the [asmap.py](https://github.com/bitcoin/bitcoin/blob/master/contrib/asmap/asmap.py) diff command, but this operates on text files within the kartograf context instead of a compressed file in Core.

I've been running adhoc scripts to do this frequently. It's needed to start comparing ASmaps continuously and analyzing the churn of networks, and it's a reasonable addition for users of Kartograf in general imo.

### DIFF
--- a/kartograf/diff.py
+++ b/kartograf/diff.py
@@ -1,0 +1,48 @@
+def diff(f1, f2, out_file):
+    '''
+    Compare two ASMap files, and write out changes to an out_file.
+    Prints the size of each type of change (assigned, reassigned, or unassigned).
+    '''
+    f1_dict = dict()
+    with open(f1, "r") as f:
+        for line in f:
+            pfx, asn = line.split()
+            f1_dict[pfx] = asn
+
+    f2_dict = dict()
+    with open(f2, "r") as f:
+        for line in f:
+            pfx, asn = line.split()
+            f2_dict[pfx] = asn
+
+    k1 = f1_dict.keys()
+    k2 = f2_dict.keys()
+
+    left = set(k1) - set(k2)
+    right = set(k2) - set(k1)
+    overlap = set(k1) & set(k2)
+
+    changes = []
+    for network in overlap:
+        asn1 = f1_dict[network]
+        asn2 = f2_dict[network]
+        if asn1 != asn2:
+            changes.append(f"{network} reassigned from {asn1} to {asn2}\n")
+
+    removed = []
+    for network in left:
+        removed.append(f"{network} unassigned from {f1_dict[network]} \n")
+
+    added = []
+    for network in right:
+        added.append(f"{network} assigned to {f2_dict[network]}\n")
+
+    with open(out_file, "w") as f:
+        f.writelines(changes + removed + added)
+        f.close()
+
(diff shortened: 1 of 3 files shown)


fjahr:
Seems good for me to add. At first I thought I would be against it because we already have this in asmap tool but you make a good point that there it is handling the encoded asmap while kartograf deals with plain text files. I agree it's valuable to add this so the files don't need to be encoded first before they can be diffed.

Ping me when you think this is ready to take out of draft status, then I can do some testing/review.

jurraca:
Updated the assigned/unassigned language to match the asmap.py a little more closely.

fjahr (commented):
Left a few code review comments but mostly I tested this against the results of `asmap-tool.py`. I think ideally the results of the two commands should be the same for compress vs. uncompressed files unless that is not possible for some reason (I can't think of any though since the compression is not supposed to be lossless).

I am not sure if there is just a different interpretation but the results don't seem to be fully consistent. I think the diff command in `asmap-tool.py` can use some improvement in terms of clarity of the output and I am not sure which side is more right as of now.

The code here reports `Changes: 12295, Removed: 25489, Added:   36124` while the result file from tool has 33979 diff lines (without `-i`). Unless I am mistaken the diff-lines should be the total of the three numbers above? I didn't go further in my analysis yet, but I will write out the commands I used below. And I used the respective files from the last two runs in `asmap-data`.

Kartograf with this PR here:

```
$ ./run diff -b=/Users/XXX/Downloads/1724248800-final_result.txt -e=/Users/XXX/Downloads/1730210400-final_result.txt
Wrote to /Users/XXX/projects/python/kartograf/out/diff.txt

Changes: 12295
Removed: 25489
Added:   36124
```

Result file: 
[diff.txt](https://github.com/user-attachments/files/17707418/diff.txt)

On `asmap-tool.py`:

```
$ contrib/asmap/asmap-tool.py diff /Users/FJ/Downloads/1724248800_asmap.dat /Users/FJ/Downloads/1730210400_asmap.dat > ~/Downloads/res.txt
```

Result file: 
[res.txt](https://github.com/user-attachments/files/17707564/res.txt)

jurraca:
So, thinking about this a little further, I think I might close this PR. The intent of the `asmap_tool.py` script is to diff the _content_ of the two ASmap files. I'm just diffing the _lines_ in the file, ignoring networks that may overlap. So the problem with the file difference is just that. I was going for an easy way to diff two ASmaps or merge files within kartograf, but the `asmap_tool` script is fully sufficient (and efficient) at doing this. It's just slightly annoying that it's not in the same repo. 

Second, the `asmap_tool.py` file does work on text files! i think I missed this. There's a small bug to fix but otherwise it works fine.

fjahr:
> Second, the `asmap_tool.py` file does work on text files! i think I missed this. There's a small bug to fix but otherwise it works fine.

Ah, ok, in that case yeah, I think then we don't need to maintain the same functionality here. Ping me when you fixed the bug :)

PR #33, opened 2024-11-05
initial CI configuration

jurraca:
Checks that the flake can be evaluated, and checks that the dev shell and package can be built.
Runs on PRs and pushes to main. 

Also updates the devShell output, which was defined in the old format.


fjahr:
ACK, looks good to me as simple start :)

PR #35, opened 2024-11-11
Introduce pylint ci task

fjahr:
This also fixes a few complaints from pylint but explicitly ignores many others that did not seem essential for the start. These can be fixed as we go along.

None of the code changes here should change behavior.

### DIFF
--- a/.github/workflows/pylint.yml
+++ b/.github/workflows/pylint.yml
@@ -1,0 +1,26 @@
+name: Pylint Check
+
+on: [push, pull_request]
+
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Check out the code
+        uses: actions/checkout@v3
+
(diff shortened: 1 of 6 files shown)


fjahr:
Checked that the ci task works (i.e. fails when it should) by reintroducing an error: https://github.com/asmap/kartograf/actions/runs/11786686736/job/32830471920

jurraca:
ACK. more an fyi, this doesn't include the pylint package in the dev environment. You'd have to split out the dev deps like in #26, so it's probably better to add it in that one or another PR.

fjahr:
> this doesn't include the pylint package in the dev environment

Yeah, but we don't use it as part of the test suite or anything like that and we don't have a lint task/script for local dev work. So as long as we only have it in CI I wouldn't put it there since the CI can do it's own setup. But we can do a linter task/script for running it locally more easily as a follow-up.

PR #36, opened 2024-11-18
add pylint configuration and package

jurraca:
pylintrc generated with `--generate-rcfile` which outputs a ton of default config. I removed the empty settings and kept the stated defaults for the rest. Our disabled rules form CI are on line 209. Would it be better to remove the default config settings?

### DIFF
--- a/pylintrc
+++ b/pylintrc
@@ -1,0 +1,417 @@
+[MAIN]
+
+# Use multiple processes to speed up Pylint. Specifying 0 will auto-detect the
+# number of processors available to use, and will cap the count on Windows to
+# avoid hangs.
+jobs=1
+
+# Control the amount of potential inferred values when inferring a single
+# object. This can help the performance when dealing with large functions or
+# complex, nested conditions.
+limit-inference-results=100
+
+# Pickle collected data for later comparisons.
+persistent=yes
+
+# Minimum Python version to use for version dependent checks. Will default to
+# the version used to run pylint.
+py-version=3.10
+
+# When enabled, pylint would attempt to guess common misconfiguration and emit
+# user-friendly hints instead of false-positive error messages.
+suggestion-mode=yes
+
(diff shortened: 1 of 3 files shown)


pylintrc
```diff
@@ -0,0 +1,417 @@
...
+        too-many-locals,
+        unspecified-encoding,
+        too-many-return-statements,
+        too-many-branches,
+        too-many-nested-blocks,
+        too-many-instance-attributes,
+        too-many-statements,
+        too-few-public-methods,
+        fixme,
+        too-many-positional,
```
fjahr, line 233: Getting this error here:

```
pylintrc:1:0: W0012: Unknown option value for '--disable', expected a valid pylint message and got 'too-many-positional' (unknown-option-value)
```

fjahr (commented):
Seems fine for me to keep the defaults in but also fine to remove them, I have no strong opinion either way.

You should also update the CI pylint command to use the rc file in this PR so that the disablements are managed in one place.

pylintrc
jurraca: what pylint version you running? Looks like nixpkgs is on 3.1.1 (fairly far behind) which has this option, but 4.0.0 has `too-many-positional-arguments` [arg](https://pylint.pycqa.org/en/latest/user_guide/messages/refactor/too-many-positional-arguments.html). 4.0.0 would be the version we should follow in this case.

jurraca: nixpkgs `24.11` release is around the corner (currently in beta) and has 3.3.1. I'll test it.

fjahr: seems like I tested this outside of the nix shell and there I have pylint 3.3.1, so yeah, using that should fix it.

jurraca: updated and tested locally with 3.3.1 :+1:

fjahr:
tACK 8dd314d6d8ecbc5dcdde03b08a4f29cf2cbb2a6f

Awesome, thanks!

PR #39, opened 2024-11-22
Remove flake-inputs; refactor and split out package

jurraca:
Besides removing a dependency for the project, this refactors the flake to be a bit cleaner.

Using `flake-utils` makes expressing "for each system" type expression simple, but it also gets in the way of items that don't need this type of expression, such as the `module.nix`. Thus we had to define an improper module expression, which makes _using_ the module in another expression more complicated as well. 

So instead we create `forAllSystems` and `nixpkgsFor` functions explicitly. This means a little more involved package expressions, but that hide less from the user. The module can now be `import`'ed and it can be used as a module in other machine configurations without extra treatment. I stole this idea from [this](https://ayats.org/blog/no-flake-utils).

I also took the opportunity to split out the main package expression into its own file, for clarity.

### DIFF
--- a/package.nix
+++ b/package.nix
@@ -1,0 +1,38 @@
+{pkgs, rpki-client }:
+let
+      pythonBuildDeps = pkgs.python311.withPackages (ps: [
+        ps.beautifulsoup4
+        ps.pandas
+        ps.requests
+        ps.tqdm
+      ]);
+in
+        pkgs.stdenv.mkDerivation {
+          # not a python-installable package, so just manually copy files
+          pname = "kartograf";
+          version = "1.0.0";
+          src = ./.;
+          nativeBuildInputs = [pkgs.makeWrapper];
+          buildInputs = [ pythonBuildDeps ];
+          propagatedBuildInputs = [ rpki-client ];
+          buildPhase = ''
+            mkdir -p $out/lib/kartograf
+            cp -r ${./kartograf}/* $out/lib/kartograf/
+          '';
+          installPhase = ''
+            mkdir -p $out/bin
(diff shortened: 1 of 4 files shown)


jurraca:
rebased. Made a nix folder for tidiness, and fixed a few things. 
This change isn't too risky imo, and I've tested it for Linux.
For other arch's, it would be good to test: 
- run `nix develop` then `python check.py`  and make sure all deps have the green checkmark.
- run `nix build` then `./result/bin/kartograf` and make sure it prints the help message

I have to think if it's worth it to add testing for the module, but for now I've tested it myself (the module issues was the reason for this change).

jurraca:
Reverted some of the module changes, which in fact did not work as expected. Tested [d5bd6cc](https://github.com/asmap/kartograf/pull/39/commits/d5bd6cca86170ad4012dc5b9e6d8076cf0eee47a) with the module pulled into a nixOS configuration and it works now.

fjahr:
@jurraca is there anyone else with better nix knowledge who could review this as well? @0xB10C could you give this a look maybe?

package.nix
```diff
@@ -0,0 +1,38 @@
...
+              --set PYTHONPATH $out/lib:$PYTHONPATH
+            wrapProgram $out/bin/kartograf \
+              --set PATH ${rpki-client}/bin:$PATH
+          '';
+          meta = with pkgs.lib; {
+            description = "Kartograf: IP to ASN mapping for everyone";
+            license = licenses.mit;
+            homepage = "https://github.com/asmap/kartograf";
+          };
+        }
```
0xB10C: nit: could format this file

PR #41, opened 2024-12-03
bump pandas requirement to >=2.2.3

jurraca:
Our current version is fairly old (released Jan 2023). This version is 3 months old and should be a safe requirement for a while. It also aligns with nixpkgs 24.11.

From discussion in #32


fjahr:
Looks good to me!

PR #42, opened 2024-12-12
Add tests for the CLI

jurraca:
In order to test the CLI, we have to be able to pass args to the parser and/or the main loop, so it makes sense to move the CLI functionality to its own module. The CLI logic is unchanged. Some formatting notices were addressed.

Includes/is stacked on #43 so we can import pytest in CI.


fjahr:
ACK

Changes look good to me and I did some light testing as well.

PR #44, opened 2024-12-13
Test the Context class

jurraca:
use pytest fixtures for the parser, update `test_cli.py` with this approach

Also adds a CLI test for empty args.

stacked on #42


fjahr:
ACK, looks good to me. Thanks!

PR #47, opened 2025-01-14
util.py: revert leading zeros logic, add is_valid_pfx

jurraca:
the previously added [check](https://github.com/asmap/kartograf/commit/97f0fbe552e9c8d4b42c8a03468f47d7bb0fb5c3) for leading zeros picks up private networks, so it's not exactly what we want here. Instead, separate the format_pfx and is_valid_pfx intents and add tests accordingly. format_pfx checks whether the existing prefix can be coerced to a string and returns it. is_valid_pfx simply checks whether it's a valid pfx or not.

Adds a test case for private networks (i.e. checks that we don't strip the leading 0, the case we missed here that led to this bug).

### DIFF
--- a/kartograf/collectors/parse.py
+++ b/kartograf/collectors/parse.py
@@ -29,7 +29,7 @@ def parse_routeviews_pfx2as(context):
                 if context.max_encode and is_out_of_encoding_range(asn, context.max_encode):
                     continue
 
-                if is_bogon_pfx(prefix) or is_bogon_asn(asn):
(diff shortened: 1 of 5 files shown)


fjahr (commented):
ACK modulo the comment, when you fix it I can merge. Thanks!

fjahr:
Will revisit the dependend PRs when this is merged :)

fjahr:
ACK

PR #48, opened 2025-01-16
sync rpki-client-nix flake for 9.4 version

jurraca:
syncs the flake input from [rpki-client-nix](https://github.com/asmap/rpki-client-nix/).

Also, makes the nixpkgs input to this flake follow the kartograf one. This just ensures that we're not evaluating two sets of nixpkgs inputs for kartograf. In other cases we would want to ensure that rpki-client retains a very specific set of inputs (for security or reproducibility reasons), and we would _not_ want to override them, but in this case it's not a particular problem for reproducibility or otherwise.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -16,22 +16,6 @@
         "type": "github"
       }
     },
-    "nixpkgs_2": {
-      "locked": {
-        "lastModified": 1727264057,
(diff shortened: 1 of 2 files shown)


fjahr:
ACK

PR #49, opened 2025-01-18
RPKI parse tests

jurraca:
Add initial tests for RPKI parsing: check that duplicates, incompletes, ROA type, and invalids are parsed correctly. Mostly, setting up a test Context and adjusting some setup.

Includes a RPKI fixture CSV and handles casting it to a `rpki_raw.json` file as expected by the parsing code.

rebased on #47


kartograf/context.py
```diff
@@ -39,6 +37,12 @@ def __init__(self, args):
             self.epoch = self.args.epoch
             self.epoch_dir = "r" + self.args.epoch
             self.epoch_datetime = repro_epoch
+        # This is only for testing, to set an epoch in the Test context.
+        # the CLI parser prevents this argument from existing without the 'reproduce' arg
+        elif self.args.epoch:
```
fjahr, line 42: Hm, couldn't we do this from outside of the context by mocking the epoch? Unless it's much more complex I would prefer not to add test specific code in the normal code.

tests/context.py
```diff
@@ -0,0 +1,51 @@
...
+            rpki_data.append(row)
+
+    output_path = os.path.join(self.out_dir_rpki, 'rpki_raw.json')
+    with open(output_path, 'w') as jsonfile:
+        json.dump(rpki_data, jsonfile, indent=2)
+
+
+def create_test_context(tmp_path, epoch):
+    TEST_ARGS.epoch = epoch
+    context = Context(TEST_ARGS)
```
fjahr: Adding to what I mentioned above, didn't test it though.

``` suggestion
            context = Context(TEST_ARGS)
            context.epoch = epoch
            context.epoch_dir = epoch
            context.epoch_datetime = datetime.utcfromtimestamp(int(epoch))
```

kartograf/context.py
```diff
@@ -48,7 +48,10 @@ def __init__(self, args):
             self.epoch_dir = str(int(utc_time_now))
             self.epoch_datetime = datetime.utcfromtimestamp(int(utc_time_now))
 
-        cwd = os.getcwd()
+        if hasattr(self.args, 'cwd'):
```
fjahr, line 51: Hm, I don't see what this is needed for.

fjahr (commented):
Took a first look but will revisit once #47 is merged.

fjahr:
Needs rebase now that #47 is merged

kartograf/context.py
jurraca: yep, had to do a small refactor to make this work here [cfb044d](https://github.com/asmap/kartograf/pull/49/commits/cfb044d0f4a20448dcb45b206053d5bc0cebb568).

kartograf/context.py
jurraca: yea this was a temporary hack around the fact that using `tmp_path` doesn't actually move the codepath, just runs the test in that tmp_path. so `os.getcwd()` was generating test data in the code path. Fixed it below in [b2a23ff](https://github.com/asmap/kartograf/pull/49/commits/b2a23ff3251a641d2f3d0d773e907e61fd32c36d).

jurraca:
I was intending to introduce just RPKI test fixtures here, but now each commit has some fairly independent changes in it:
- changing f-strings to `os.path.join` should be harmless, just prevents us from having to manually set `/` path delimiters when creating paths.
- [cfb044d](https://github.com/asmap/kartograf/pull/49/commits/cfb044d0f4a20448dcb45b206053d5bc0cebb568) includes the RPKI fixtures and tests that check the parsing for those fixtures. In order to do so, it introduces a test Context in `tests/context.py`. It also modifies the `epoch` handling in the parsing code, in order to be able to mock it properly in this test context. (let me know if it makes sense to split this up).
- [b2a23ff](https://github.com/asmap/kartograf/pull/49/commits/b2a23ff3251a641d2f3d0d773e907e61fd32c36d) fixes the path in the test context to properly create test data in the `tmp_path`.

kartograf/context.py
```diff
@@ -9,14 +9,14 @@ class Context:
     def __init__(self, args):
         self.args = args
 
-        # The epoch is used to keep artifacts seperated for each run. This
+        # The epoch is used to keep artifacts separated for each run. This
         # makes cleanup and debugging easier.
         if args.wait:
```
fjahr, line 14: nit: Could update this to `self.args.wait`

kartograf/context.py
```diff
@@ -51,21 +40,22 @@ def __init__(self, args):
...
-        self.out_dir_irr = f"{self.out_dir}irr/"
-        self.out_dir_rpki = f"{self.out_dir}rpki/"
-        self.out_dir_collectors = f"{self.out_dir}collectors/"
+            self.data_dir = os.path.join(cwd, "data", self.epoch_dir)
 
         if os.path.exists(self.data_dir) and not self.reproduce:
             print("Not so fast, a folder with that epoch already exists.")
             sys.exit()
 
+        self.data_dir_irr = os.path.join(self.data_dir, "irr")
```
fjahr, line 49: Instead of os.path.join could use Pathlib here too

fjahr (commented):
Looks pretty good, leaving a few comments that are mostly non-essential but the util file should be moved to the folder and renamed.

Don't forget to take it out of draft when you think it could be ready to merge.

jurraca:
I moved the pathlib / os.path changes to #54 for clarity, this PR is rebase on that, so let's merge that first. still many follow ups to be had with this, mainly in terms of fixtures and parsing scenarios...

jurraca:
do not merge yet -- found a bug

jurraca:
rebased on #56 , which should go first.

fjahr:
nit on commit structure: I would have preferred if you had squashed the last two commits into the previous ones so that that the same lines/files are not changed twice and that you had split the context refactoring in the second commit into it's own commit.

fjahr (commented):
tACK

PR #50, opened 2025-01-21
bug fix: checking reproduce and epoch args logic

jurraca:
A bug I noticed while testing: the check that both `reproduce` and `epoch` args only occur together is broken since this [commit](https://github.com/asmap/kartograf/commit/1b6bc2136dc0efca9d87652a915290cf709341d6). The commit intended to ensure that we print the help text when no arguments were provided, but it did not include any tests, which was a mistake on my part and would have caught this. 
The `command` to check here is `map` not `run`. The logical operator was also wrong for the intended check.
The test checks the lack of both arguments, and that a helpful usage text is show to the user.

### DIFF
--- a/kartograf/cli.py
+++ b/kartograf/cli.py
@@ -81,8 +81,8 @@ def main(args=None):
     parser = create_parser()
     args = parser.parse_args(args)
 
-    if args.command == "run":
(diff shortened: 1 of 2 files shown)


fjahr:
utACK, code and test makes sense.

PR #51, opened 2025-01-22
merge_test: use networks sets instead of counts, updated comments

jurraca:
follow-ups from #37  : use actuals instead of counts of entries.


tests/merge_test.py
```diff
@@ -50,9 +52,9 @@ def test_merge_from_fixtures(tmp_path):
     general_merge(base_path, extra_path, None, outpath)
     with open(outpath, "r") as f:
         l = f.readlines()
-        final_network_count = len(l)
-        expected_count = (len(base_nets) + len(extra_nets)) - (base_subnet_count + extra_subnet_count)
-        assert final_network_count == expected_count
+        final_networks = {line.split()[0] for line in l}
+        expected_networks = set(base_nets + extra_nets) - set(base_subnets + extra_subnets)
```
fjahr, line 56: Hm, why do we need the `set` here? Doesn't the merge code handle duplicates as well?

tests/merge_test.py
```diff
@@ -17,30 +17,32 @@ def __tmp_paths(tmp_path):
...
 def test_merge_from_fixtures(tmp_path):
     '''
-    Assert that general_merge merges subnets correctly.
+    Assert that general_merge merges subnets correctly,
+    and validates against expected network sets, i.e. invalid networks are not merged,
+    and subnets are merged into the root network appropriately.
     '''
 
-    base_nets, base_subnet_count = __read_test_vectors("tests/data/base_file.csv")
+    base_nets, base_subnets = __read_test_vectors("tests/data/base_file.csv")
```
fjahr, line 43: Hm, should (here and elsewhere in the file) `subnet` be renamed to something more general like `exludeds` or so? Because it also includes invalids that are excluded even if they may not be a subnet. And also when we have duplicates they are also not included, right?

fjahr (commented):
Cool, even simpler than I expected :) Should be almost good to go but please check if my comments make sense first.

tests/merge_test.py
fjahr: And if duplicates is not a case we cover here yet, I would suggest you can just add it basically as another variant of the subnets test cases.

tests/merge_test.py
jurraca: :+1: done

fjahr (commented):
I think I would like to simplify this a bit more and I have sketched it out here: https://github.com/asmap/kartograf/commit/37ce08995d9adecd555bd31caf87c81ae35982cd

This key pieces here (this echoes my comments which I wrote before, so you can mostly ignore them):
- I don't think we need to distinguish between invalid and subnet/dublicate in the fixture because the merge function doesn't handle invalids. So we also don't need to test this. Instead just mark each entry if it should be in the result or not
- Resolve `is_valid` vs `test_case` and turn `test_case` into comment for documentation only
- Don't use `set` because I it can hide test failures by sorting out networks we didn't want to see included, just sort so we can compare the lists 1-to-1
- The other (style) changes are not super relevant

tests/merge_test.py
fjahr: I think this is more complicated than it has to be. I sketched out below what I was thinking of since that's clearer than describing it :) Let me know if it makes sense.

tests/test_merge.py
```diff
@@ -41,9 +41,9 @@ def test_merge_from_fixtures(tmp_path):
...
     and subnets are merged into the root network appropriately.
     '''
-    base_nets, base_subnets = __read_test_vectors("tests/data/base_file.csv")
+    base_nets, base_nets_to_exclude = __read_test_vectors("tests/data/base_file.csv")
     base_path = tmp_path / "base.txt"
-    extra_nets, extra_subnets = __read_test_vectors("tests/data/extra_file.csv")
+    extra_nets, extra_nets_to_exclude = __read_test_vectors("tests/data/extra_file.csv")
     extra_path = tmp_path / "extra.txt"
     # write the networks to disk, generating ASNs for each network
     generate_ip_file(base_path, build_file_lines(base_nets, generate_asns(len(base_nets))))
```
fjahr, line 50: Not sure if I mentioned this before but as a next step we'll want to make the ASNs part of the fixture files too since they also allow us to test the scenarios where the ASN is relevant for the network being included or not. But let's keep this for a follow-up.

jurraca:
thanks, that indeed makes more sense and is simpler. I didn't see anything else to improve for that specific test, did I miss something? The other tests will be updated in a follow up with the ASN thing.

Regarding duplicates, I missed that and you're right we don't merge duplicates since we don't expect them. Do you think we should add logic to filter them out if we see them (likely in the sort step) ?

fjahr:
> Regarding duplicates, I missed that and you're right we don't merge duplicates since we don't expect them. Do you think we should add logic to filter them out if we see them (likely in the sort step) ?

I don't think we need to add that unless a need arises. I think it's good that we have clear boundaries: The merge code just merges stuff and ignores internal issues in the inputs. The parse code handles quality issues that are internal to the inputs. This makes sense to me but I also came up with it I guess :D Most importantly I don't want to repeat the same thing in two places unnecessarily because that can lead to other types of confusion.

fjahr:
ACK

PR #52, opened 2025-01-22
tests: align test file names, relative path fixup

jurraca:
I was mixing test file naming schemes, this aligns on `tests/test_*.py`.
Also, handle test data relative paths correctly.


tests/test_merge.py
```diff
@@ -37,10 +39,10 @@ def test_merge_from_fixtures(tmp_path):
     '''
     Assert that general_merge merges subnets correctly.
     '''
-
-    base_nets, base_subnet_count = __read_test_vectors("tests/data/base_file.csv")
+    data_dir = os.path.join(os.path.dirname(__file__), "data")
```
fjahr, line 42: nit: Instead of `os.path.join` I would prefer if you would continue using Pathlib and the `/` operator like in the rest of the code here where paths are concatenated.

fjahr (commented):
Looks good the me aside from the style nit. I will just merge this and maybe you can keep my comment in mind for the future when you touch that file again.

PR #53, opened 2025-01-22
CI: run python tests in new job

jurraca:
Adds a new job to run tests on push and PR.
Also, fixes existing build outputs job (wasn't running since it was pointed to the wrong branch).


fjahr:
ACK

Just FYI you can link to failed CI cases later instead of using a screenshot: https://github.com/asmap/kartograf/actions/runs/12907465109/job/35990942993 You can find the link at the top of the old commit page.

PR #54, opened 2025-01-30
context: use pathlib instead of os.path

jurraca:
f-strings require getting the path slashes correct manually, so pathlib is better here, albeit more verbose and having to cast paths back to strings. Adds isinstance tests for context fields. 

There's more path instances to be fixed elsewhere but this is a first pass done in handling #49 .

### DIFF
--- a/kartograf/context.py
+++ b/kartograf/context.py
@@ -1,5 +1,5 @@
 from datetime import datetime
-import os
+from pathlib import Path
 import sys
 import time
 
@@ -20,12 +20,12 @@ class Context:
 
         self.reproduce = self.args.reproduce is not None
         if self.reproduce:
-            items = os.listdir(self.args.reproduce)
+            items = Path.iterdir(Path(self.args.reproduce))
             source_folders = []
             for folder in items:
-                full_path = os.path.join(self.args.reproduce, folder)
-                if os.path.isdir(full_path):
-                    source_folders.append(folder)
+                full_path = self.args.reproduce / folder
+                if Path.is_dir(full_path):
+                    source_folders.append(folder.name)
             # We override the args because we are reproducing and only where we
             # have data we try to use is, the actual args passed don't matter.
             self.args.irr = 'irr' in source_folders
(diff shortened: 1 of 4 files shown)


kartograf/context.py
```diff
@@ -44,47 +44,52 @@ def __init__(self, args):
...
-        self.data_dir_collectors = f"{self.data_dir}collectors/"
+            self.data_dir = str(cwd / "data" / self.epoch_dir)
+
+        if Path(self.data_dir).exists() and not self.reproduce:
+            print("Not so fast, a folder with that epoch already exists.")
+            sys.exit()
+
+        self.data_dir_irr = str(Path(self.data_dir) / "irr")
+        self.data_dir_rpki_cache = str(Path(self.data_dir) / "rpki/cache")
+        self.data_dir_rpki_tals = str(Path(self.data_dir) / "rpki/tals")
```
fjahr, line 62: Could you split these as well?

```suggestion
        self.data_dir_rpki_cache = str(Path(self.data_dir) / "rpki" / "cache")
        self.data_dir_rpki_tals = str(Path(self.data_dir) / "rpki" / "tals")
```

fjahr (commented):
Great, looks like a good improvement to me. If you can split the leftover strings containing I think this is ready for merge.

tests/test_merge.py
```diff
@@ -1,5 +1,6 @@
 '''
 Test merging multiple sets of networks, as if they were independent AS files.
+<<<<<<< HEAD
```
fjahr, line 3: Leftover from a merge conflict here

tests/test_merge.py
```diff
@@ -40,10 +41,10 @@ def test_merge_from_fixtures(tmp_path):
     and validates against expected networks,
     i.e. subnets are merged into the root network appropriately.
     '''
-    testdir = Path(__file__).parent
-    base_nets, base_results = __read_test_vectors(testdir / "data/base_file.csv")
+    data_dir = Path(__file__).parent / "data"
+    base_nets, base_subnet_count = __read_test_vectors(data_dir / "base_file.csv")
```
fjahr, line 45: There are still the old variable names here. Probably a silent merge conflict.

jurraca: sorry :disappointed:, fixed.

fjahr:
ACK

PR #55, opened 2025-02-03
Add tests for RPKI ASN tiebreaks

jurraca:
rebased on #49 . 

As a result of adding more entries to the fixtures, updates the duplicate and entries counts in the `test_roa_validations` test. Also it removes the "count of duplicates" assertion which wasn't meaningful.

~Also, and this should really be a separate commit, but~ edited commits, I used simpler fixtures for these new tests, and so went back and fixed the initial set of fixtures to match those. The fixtures CSV is now more readable.


tests/test_rpki_parser.py
```diff
@@ -96,3 +96,48 @@ def test_roa_incompletes(tmp_path, capsys):
...
+def test_roa_valid_until_fallback(tmp_path):
+    '''Test ROA selection falls back to later valid_until'''
+    epoch = "111111111"
+    context = create_test_context(tmp_path, epoch)
+    setup_test_data(context)
+    parse_rpki(context)
+
+    final_path = os.path.join(context.out_dir_rpki, "rpki_final.txt")
+    with open(final_path, "r") as f:
+        entries = [line.strip() for line in f.readlines()]
```
fjahr, line 100: Nit: Quite a bit of dublication between the newly added tests. Maybe you can extract this into a helper function that is local in this file?

fjahr: Meh, I will just merge this since it looks fine otherwise. Maybe keep this in mind for the next time you add something to this file :)

PR #56, opened 2025-02-03
Bugfix for path concatenation

jurraca:
The [latest commit](https://github.com/asmap/kartograf/commit/b6c630eeb9177dd2bd2b14d5e685e483be562fa8) on master (using pathlib to do path concatenation in the context) breaks every other path concatenation step in other modules. 

I should have done this in one go. Unfortunately, better tests would only have caught some of these issues. The context test is updated here, but we'll need to think about how to test other file handling during the fetching and merge, since file generation and lookup happen outside the context, for example in `collectors/routeviews.py`. This PR should cover all the instances of path concatenation for files in the app though.

I only caught this by doing a full run, and I'll do that for each change moving forward until we have more tests in place.

### DIFF
--- a/kartograf/collectors/parse.py
+++ b/kartograf/collectors/parse.py
@@ -1,3 +1,5 @@
+from pathlib import Path
+
 from kartograf.bogon import (
     is_bogon_pfx,
     is_bogon_asn,
@@ -9,8 +11,8 @@ from kartograf.util import parse_pfx
 
 @timed
 def parse_routeviews_pfx2as(context):
-    raw_file = f'{context.out_dir_collectors}pfx2asn.txt'
-    clean_file = f'{context.out_dir_collectors}pfx2asn_clean.txt'
+    raw_file = Path(context.out_dir_collectors) / "pfx2asn.txt"
+    clean_file = Path(context.out_dir_collectors) / "pfx2asn_clean.txt"
 
     print("Cleaning " + raw_file)
     with open(raw_file, 'r') as raw, open(clean_file, 'w') as clean:
--- a/kartograf/collectors/parse.py
+++ b/kartograf/collectors/parse.py
@@ -14,7 +14,7 @@ def parse_routeviews_pfx2as(context):
     raw_file = Path(context.out_dir_collectors) / "pfx2asn.txt"
     clean_file = Path(context.out_dir_collectors) / "pfx2asn_clean.txt"
 
-    print("Cleaning " + raw_file)
+    print("Cleaning " + str(raw_file))
     with open(raw_file, 'r') as raw, open(clean_file, 'w') as clean:
         lines = raw.readlines()
         for line in lines:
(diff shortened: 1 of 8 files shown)


kartograf/collectors/parse.py
```diff
@@ -9,8 +11,8 @@
 
 @timed
 def parse_routeviews_pfx2as(context):
-    raw_file = f'{context.out_dir_collectors}pfx2asn.txt'
-    clean_file = f'{context.out_dir_collectors}pfx2asn_clean.txt'
+    raw_file = Path(context.out_dir_collectors) / "pfx2asn.txt"
+    clean_file = Path(context.out_dir_collectors) / "pfx2asn_clean.txt"
 
     print("Cleaning " + raw_file)
```
fjahr, line 17: Got a failure here

```
Traceback (most recent call last):
  File "/Users/FJ/projects/python/kartograf/./run", line 11, in <module>
    main()
  File "/Users/FJ/projects/python/kartograf/kartograf/cli.py", line 94, in main
    Kartograf.map(args)
  File "/Users/FJ/projects/python/kartograf/kartograf/kartograf.py", line 88, in map
    parse_routeviews_pfx2as(context)
  File "/Users/FJ/projects/python/kartograf/kartograf/timed.py", line 10, in wrapper
    result = func(*args, **kwargs)
             ^^^^^^^^^^^^^^^^^^^^^
  File "/Users/FJ/projects/python/kartograf/kartograf/collectors/parse.py", line 17, in parse_routeviews_pfx2as
    print("Cleaning " + raw_file)
          ~~~~~~~~~~~~^~~~~~~~~~
TypeError: can only concatenate str (not "PosixPath") to str
```

Probably just needs this:

```suggestion
    print("Cleaning " + str(raw_file))
```

fjahr:
> I only caught this by doing a full run, and I'll do that for each change moving forward until we have more tests in place.

Yeah, we should keep doing it in the foreseeable future. I've been doing it for the bigger change almost every time but didn't repeat it always when you addressed feedback, that's probably how it slipped through. It's annoying but a good motivator to get better test coverage.

What do you think about a functional end-to-end test where it's just an reproduction run with some trimmed RPKI, IRR and RV data and we just require it to not crash? We wouldn't even need to keep the trimmed input data in the repository, we could have an external repository which the CI job then clones and runs the repo on.

jurraca:
> a functional end-to-end test where it's just an reproduction run with some trimmed RPKI, IRR and RV data 

yea was thinking about this and definitely worth doing. I'll set it up :+1: It would have caught this for sure.

jurraca:
ok successful run based on [e85cf4c](https://github.com/asmap/kartograf/pull/56/commits/e85cf4cbad404d48c75fa8f88649088aaf3f2109)

fjahr:
tested ACK e85cf4cbad404d48c75fa8f88649088aaf3f2109

fjahr:
> yea was thinking about this and definitely worth doing. I'll set it up 👍 It would have caught this for sure.

Awesome! Should give us some more peace of mind/early feedback.

PR #58, opened 2025-02-08
IRR parsing tests: add fixtures and basic test, update context

jurraca:
Initial PR for IRR parsing tests. Mostly setup, keeping the PR small.

Adds an IRR file to use as fixtures, and updates the test context to use them.
Adds a basic parsing test. 

branched off of #49 in order to use the test context, so this should go after.

### DIFF
--- a/tests/test_util.py
+++ b/tests/test_util.py
@@ -1,6 +1,6 @@
+import pytest
 from kartograf.util import parse_pfx, is_valid_pfx, get_root_network, rir_from_str
 
-import pytest
 
 def test_valid_ipv4_network():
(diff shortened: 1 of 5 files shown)


tests/test_util.py
```diff
@@ -1,6 +1,6 @@
+import pytest
 from kartograf.util import parse_pfx, is_valid_pfx, get_root_network, rir_from_str
 
-import pytest
```
fjahr: nit: belonged in the previous commit

fjahr (commented):
ACK

looks all good to me

PR #59, opened 2025-02-09
Update print format to match removal of progress bar

fjahr:
Minor format improvement. I think after that print statement we used to have the progress bar until recently. Now that it's gone the colon and extra newline don't make sense, so we can remove them.

### DIFF
--- a/kartograf/merge.py
+++ b/kartograf/merge.py
@@ -146,7 +146,7 @@ def general_merge(
     print("Parse extra file to Pandas DataFrame")
     df_extra = extra_file_to_df(extra_file)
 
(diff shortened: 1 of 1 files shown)


fjahr:
See the output in my latest test run that caught my eye:

```
Parse base file to dictionary
Parse extra file to Pandas DataFrame
Merging extra prefixes that were not included in the base file:

Finished merging extra prefixes.
Finished filtering! Originally 1258502 entries filtered down to 485798
Merging base file with filtered extra file
```

jurraca (approved):
ACK

PR #60, opened 2025-02-10
Ci mock run

jurraca:
Adding a mock, end-to-end run of the `map` process.

I would like to use actual downloaded RPKI data instead of a already validated data which would require a new flag just for this. However, having issues using actual RPKI data. I added two repositories: 
- a subset of the `rpki.arin.net` repo, with intermediate certs, CRL and MFT files.
- a full set from `rpki.sunoaki.net`. 

All these end up as `invalid` with the error `cannot find a local certificate issuer`. I think I have some gaps in understanding process here. 
- the `rpki-client` validation enforces the offline flag, so it should look for certs locally.
- The certs are in the repository path, and afaict `rpki-client` looks for the certs there. 
- the certificates are valid, with their `valid_until` date set far into the future.
- I'm not sure if you need the full recursive path -- with the (nested) ARIN repo, it looks like yes, which is why i added the whole path and chain of certs, but for sunoaki this should not be an issue. All end up as invalid.

Curious if you have an idea @fjahr if not I'll ask the rpki-client team. 
Thanks.

### DIFF
--- a/.github/workflows/test-reproduction-run.yml
+++ b/.github/workflows/test-reproduction-run.yml
@@ -25,12 +25,5 @@ jobs:
         path: tests/data
 
     - name: Run with reproduction data
-      id: run_cmd
       run: |
-        output=$(nix develop --command ./run map -r tests/data/1742127147 -t 1742127147)
-        echo "output=$output" >> $GITHUB_OUTPUT
-
-    - name: Check output hash
-      if: ${{join(steps.run_cmd.outputs.*, '') contains '76835fb5d0002930ede434f396b38ab690e2b38193a516a5fb513ad2aa22e5af'
-      run: |
-        echo "Output matches the expected value"
+        nix develop --command ./run map -r tests/data/1742127147 -t 1742127147
--- a/.github/workflows/build-nix-outputs.yml
+++ b/.github/workflows/build-nix-outputs.yml
@@ -1,7 +1,9 @@
-name: Build outputs
+name: Build Outputs
 
 on:
   push:
+    branches: [master]
+  pull_request:
     branches: [master]
 
 jobs:


fjahr:
I not a big fan of adding all these files to the repository, it's pretty noisy. Can't we add a dummy reproduction data repository that the CI clones and uses instead? I suggested this previously but not sure if you overlooked it.

I think I didn't see Routeviews data in here yet, I would like to have that too so that code is covered as well. I think the easiest way to do this is to get an actual reproduction run, trim the data of each of the data sources by deleting parts of it (a lot). Then let the CI run a normal reproduction run with the same timestamp and evaluate the hash result to be what we expect.

I think the data in a separate repo can be valuable in other contexts too, like doing a workshop where you can't be sure the participants have a lot of time and online connectivity. We could also add more reproduction runs to the repo that hit different edge cases without poluting the repo here with thousands of files.

For the error you describe: You think you removed the `/ta` folder which includes the actual trust anchors. Without these `rpki-client` can't validate the ROAs and since the repro run is offline it doesn't try to use the TALs to download them again.

.github/workflows/test-reproduction-run.yml
```diff
@@ -0,0 +1,23 @@
...
+      with:
+        python-version: '3.11'
+
+    - name: Install dependencies
+      run: |
+        pip install -r requirements.txt
+
+    - name: Run with reproduction data
+      run: |
+        ./run map -r tests/data/ci-run -t $(date +%s)
```
fjahr, line 23: As mentioned in the other message, I would prefer if we would run a reproduction run with the actual static timestamp and then evaluate the hash to be what we expect. With the dynamic date the hash wouldn't be deterministic and I think we might be missing on some potential coverage of our deteriminism.

fjahr:
> I'll close this PR either way but I wanted to validate my understanding first.

No need to close, the ci job will need to be added to this repo anyway. Just move the data out to a seperate repo and keep this for the CI job so we can continue the discussion here.

> I'm still having trouble making this work. Unless I'm misunderstanding, the cert revocation lists (CRL) have an expiry, and so any validation will fail if the CRL is expired. Every time I try to validate an ROA with the full cache (including the /ta directory), it fails. If so, we can't use a set of fixtures unless we create our own dummy ones with expiry dates which are far into the future.

Sounds strange. From when is that data? I don't have issues with reproducing a run from 6 months ago for example so IMO the CRLs should also respect our -t arg. Can you push the full data you are using? Or maybe add them to the separate repo right away. Using a pruned cache should work, I have done it for a workshop in the past but it wasn't straight forward to create. There is probably still something very subtle missing.

Oh, and do you really mean the full cache, i.e. a full reproduction run or the pruned version here?

I can give it a shot to prune a cache myself as well but I won't get to it before next week.

jurraca:
I'm using pruned data from an RPKI data set. Assuming the data in `tests/data/ci-run` is pruned data from epoch `1739359651` I'm running `./run map -r tests/data/ci-run -t 1739359651`. Even with a static date or a dynamic date, or hard coding it to the source timestamp, the reproduce run results in a 100% invalid ROAs and therefore no data in the output file.

The `tests/data/ci-run` have the `ta/` folder of self-signed certs for each RIR for that run.

<details> <summary>Example result</summary>
--- Start Kartograf ---

Kartograf version: 0.4.8
Using rpki-client version 9.4 (recommended).
The epoch for this run is: 1739359651 (2025-02-12 11:27:31 UTC, local: 2025-02-12 11:27:31 WET)
This is a reproduction run based on the data in tests/data/ci-run/

--- Validating RPKI ---

Validating RPKI ROAs
76 raw RKPI ROA files found.
100%|██████████████████████████████████████████████████████| 76/76 [00:00<00:00, 2296.31it/s]
76 RKPI ROAs validated and saved to /home/base/code/asmap/kartograf/out/r1739359651/rpki/rpki_raw.json, file hash: 396f0d4318ff91c5af490a064a36ce518577f37638e5625b0e4f0f779cf4314e
...finished in 0:00:00.075559

--- Parsing RPKI ---

Parsing 76 ROAs
Result entries written: 0
Duplicates found: 0
Invalids found: 76
Incompletes: 0
Non-ROA files: 0
...finished in 0:00:00.000610

--- Sorting results ---

...finished in 0:00:00.000229

--- Finishing Kartograf ---

The SHA-256 hash of the result file is: e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855
Total runtime: 0:00:00.077218
</details>

I can also run a full reproduction job from older dates without any issue, so I think something else is missing from the certificate chain that allows `rpki-client` to validate (i.e. we're "pruning wrong"). I'll try with some other subsets of the rpki data.

I've created a repo and put a fresh set of data [here](https://github.com/asmap/sample-data).

fjahr:
I have played around with the data today and opened a PR to the sample repo: https://github.com/asmap/sample-data/pull/1

Let me know if that one works for you. As I wrote there, there does seem to be some kind of interdependency between the RIR and non-RIR repos that we don't understand yet. Would definitely good to know what we are missing there.

fjahr:
Ok, some further research suggests that this is basically the hierarchical structure we see playing out here. Even if we have the TAs and Certs for the RIR repos externally the RIR repositories probably contain trust anchors and certificates issued to their subordinates. The non-IRR repos depend on these parent certificates from the RIR repos.

See, for example here: https://labs.ripe.net/author/gih/models-of-trust-for-the-rpki/ but please challenge this if you want to do some further digging!

fjahr:
Hm, strange that the newly added check doesn't show up in the list of checks here in the bottom of the PR. Do you know why that is?

jurraca:
doesn't CI config only kick in once merged?

fjahr:
I don't think so, I tested the pylint CI task for example before merging by introducing an error and saw the CI task fail within the PR: https://github.com/asmap/kartograf/pull/35 It makes sense that the new check would run here because the checks run on the code in the PR so it should pick up the new check, too, I think...

.github/workflows/test-reproduction-run.yml
```diff
@@ -0,0 +1,30 @@
+name: Test Full Run
+
+on:
+  push:
+    branches: [master]
```
fjahr, line 5: The issue is here, this only runs on master. It should run on the same triggers as the other CI jobs.

fjahr:
I've noticed that the nix build outputs CI job also doesn't run here, maybe you can fix that as well while you are at it. Maybe also rename the file, it's currently just called `ci.yml` which confused me :)

jurraca:
OK, the reproduction run now runs, and renamed the Nix outputs job :+1:

fjahr:
Ok, I have merged the PR in `sample-data`, can you adapt the CI to use master branch instead of the PR?

fjahr:
And can you look into matching the final hash that we expect as the last step of the CI? This way we would not only detect if the reproduction errors but also if we have unwanted change in the resulting final hash...

fjahr:
If it's too complicated we can leave it for a follow-up though. Only if gh actions provides a simple "assert that string is in log outputs" or something.

jurraca:
> can you adapt the CI to use master branch instead of the PR?

Ah, it's already pointed at main, not the PR's branch. So it tested against the rpki data but not the PR data. will trigger a run manually.

> And can you look into matching the final hash that we expect as the last step of the CI

yep, looking

.github/workflows/test-reproduction-run.yml
```diff
@@ -0,0 +1,29 @@
...
+    - name: Checkout Sample Data
+      uses: actions/checkout@v4
+      with:
+        repository: asmap/sample-data
+        ref: main
+        path: tests/data
+
+    - name: Run with reproduction data
+      run: |
+        nix develop --command ./run map -r tests/data -t 1742127147
```
fjahr, line 29: I see, since I kept the data in the timestamp folder in my PR I think you will need to change this here still

```suggestion
nix develop --command ./run map -r tests/data/1742127147 -t 1742127147
```

jurraca:
I can't seem to get a good output check working, will return to it. The last two runs ran successfully with the `sample-data` master branch.

fjahr:
ACK

Very nice, the run takes 4 minutes which is roughly what I was hoping for. Not instant but that should hardly slow development down.

PR #61, opened 2025-02-18
bugfix: handle reproduce directory paths correctly

jurraca:
The 'full_path' variable in the reproduce step is incorrect, creating non-existent paths. The test passes because the path construction in kartograf.context checks that the name of the directory is valid but doesn't check that the path was constructed correctly. So the test passes, but the reproduce command fails.
This commit updates the path construction logic and adds an assertion that the path in the context exists.

### DIFF
--- a/kartograf/context.py
+++ b/kartograf/context.py
@@ -20,12 +20,9 @@ class Context:
 
         self.reproduce = self.args.reproduce is not None
         if self.reproduce:
-            items = Path.iterdir(Path(self.args.reproduce))
-            source_folders = []
-            for folder in items:
-                full_path = self.args.reproduce / folder
-                if Path.is_dir(full_path):
-                    source_folders.append(folder.name)
+            data_path = Path(self.args.reproduce)
+            # in the data path, get the names of directories in the given path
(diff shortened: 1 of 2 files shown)


fjahr:
Hm, I tried a reproduction with master and I didn't see a failure. I also printed out the `full_path` variable and it is correct. The test also doesn't fail for me on master. So I am not sure this is actually a bug. We can still merge it, certainly the test. But would be good to understand why you it fails for you, can you check again that this is indeed still the case for you?

fjahr:
ACK

jurraca:
thanks -- I just ran again and it didn't blow up as well. I ran into a bug while running reproduce and I thought this was the problem, and I ran this locally again to double check: 
```
>>> from pathlib import Path
>>> items = Path.iterdir(Path("data/1739816170"))
>>> for folder in items:
...     "data/1739816170" / folder
...
PosixPath('data/1739816170/data/1739816170/rpki')
PosixPath('data/1739816170/data/1739816170/collectors')
PosixPath('data/1739816170/data/1739816170/irr')
```

so I concluded that the `Path.iterdir` was returning the full path and we were appending to it. But it seems that's not the case, and I'm still confused. So yea I would argue this is still a good change since it's clearer.

PR #65, opened 2025-03-21
test: add check for invalid "--wait" arg

jurraca:
checks that the argument is not in the past.
This checks it at the CLI parsing level. Context tests have past timestamps for example, and this is correct since the context does not check the validity of the args.

### DIFF
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -41,6 +41,13 @@ def test_map_with_options(parser):
     assert args.reproduce == '/path'
     assert args.epoch == '123'
 
+def test_map_with_past_wait(capsys):
+    args = ['map', '-w', '1225411200']
+    with pytest.raises(SystemExit):
+        main(args)
+    captured = capsys.readouterr()
+    assert captured.err.startswith("usage:")
+
 def test_merge_command(parser):
     args = parser.parse_args(['merge'])
     assert args.command == 'merge'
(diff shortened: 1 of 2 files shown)


tests/test_cli.py
```diff
@@ -41,6 +41,13 @@ def test_map_with_options(parser):
     assert args.reproduce == '/path'
     assert args.epoch == '123'
 
+def test_map_with_past_wait(capsys):
+    args = ['map', '-w', '1225411200']
+    with pytest.raises(SystemExit):
+        main(args)
+    captured = capsys.readouterr()
+    assert captured.err.startswith("usage:")
```
fjahr, line 49: Would prefer if you could make this more precise:

```suggestion
    assert "Cannot wait for a timestamp in the past (1225411200)" in captured.err
```

fjahr: We have this pattern with `captured.err.startswith("usage:")` in a few other places in the test, I would prefer if the actual error message could be identified in the check in these other places too but that can be further improved in a follow-up if you don't want to do it here (I can do it too). This should prevent that checks are not catching errors if a check becomes unreachable or is hidden behind another error.

fjahr (commented):
Looks good to me but would prefer the tighter check on the error message.

tests/test_cli.py
jurraca: thanks yea i'm not sure why I left it like this. fixed other tests where appropriate below.

fjahr:
ACK, thanks for changing the other checks too

PR #67, opened 2025-03-22
feat: check if kartograf is being run as root

### LINKED ISSUE #66: Throw error when running as root
As pointed out [here](https://github.com/asmap/asmap-data/issues/22#issuecomment-2744040614), rpki-client will not work as expected when run as root. Kartograf should attempt to detect when run as root, and throw an error.

luisschwab:
Closes #66.

This PR adds a security check to prevent Kartograf from running with root privileges. If so, it exits with the message `Kartograf must be ran with a non-root user.`

### DIFF
--- a/kartograf/cli.py
+++ b/kartograf/cli.py
@@ -76,10 +76,16 @@ def create_parser():
 
     return parser
 
+def is_root():
+    # Get current the Effective UID. On Unix systems the root user EUID is 0.
+    return os.geteuid() == 0
 
 def main(args=None):
     parser = create_parser()
     args = parser.parse_args(args)
+
+    if is_root():
+        sys.exit("Kartograf must be run as a non-root user.")
 
     if args.command == "map":
         if not args.reproduce and args.epoch:


kartograf/cli.py
```diff
@@ -76,11 +76,17 @@ def create_parser():
 
     return parser
 
+def is_root():
+    # Get current the Effective UID. On Unix systems the root user EUID is 0.
+    return os.geteuid() == 0
```
fjahr, line 81: There seems to be a tab at the end of this line

kartograf/cli.py
```diff
@@ -76,11 +76,17 @@ def create_parser():
...
+def is_root():
+    # Get current the Effective UID. On Unix systems the root user EUID is 0.
+    return os.geteuid() == 0    
 
 def main(args=None):
     parser = create_parser()
     args = parser.parse_args(args)
 
+    if is_root():
+        sys.exit("Kartograf must be ran with a non-root user.")
```
fjahr, line 88: nit: I think that's more correct but I'm not a native speaker :)

```suggestion
        sys.exit("Kartograf must be run as a non-root user.")
```

fjahr (commented):
Tested and it looks good to me, would be nice if you could address the tab though

luisschwab:
Fixed!

kartograf/cli.py
```diff
@@ -76,11 +76,17 @@ def create_parser():
 
     return parser
 
+def is_root():
+    # Get current the Effective UID. On Unix systems the root user EUID is 0.
+    return os.geteuid() == 0
```
fjahr, line 81: Hm, now there is a space at the end. Sorry for being annoying but my editor screams at me about this :) We should probably configure the CI to detect this.

luisschwab: My bad, need to configure mine to do the same

fjahr:
ACK 56c72c7763d634c930dcdf54b0354e5156b1cf81

Code looks good to me and from my research this seems to be a good solution for all the systems we care about primarily. It probably doesn't work for windows but I don't think we have had a windows user yet and we already mention in the readme that it is untested on Windows so I think it's acceptable to ignore it for now.

luisschwab:
> It probably doesn't work for windows but I don't think we have had a windows user yet and we already mention in the readme that it is untested on Windows so I think it's acceptable to ignore it for now.

Yes, I took this into consideration when making this PR.

PR #70, opened 2025-03-28
rpki: Rely on rpki-client management of TALs again

fjahr:
This change depends on the upcoming `rpki-client` release (9.5) which isn't out yet, that's why it is in draft. Not sure when it would be a good time to merge, we already ask people to use the latest `rpki-client` but 9.5 would need to be widely available at least and this is also not critical so we can probably wait with this for a while.

We are currently managing the TALs ourselves, download them and give them to `rpki-client` as options. We had to do this because `rpki-client` didn't include the ARIN TAL in their releases for a while because of some licensing concerns: https://github.com/rpki-client/rpki-client-portable/issues/80 However, in January ARIN revised the license and so now all TALs will be included in the releases again: https://github.com/rpki-client/rpki-client-portable/commit/05de5a20a69f30a7d99ee2bd24d2512f28dc158f

That means we can remove a bunch of code, when we feel ready for it.

### DIFF
--- a/kartograf/context.py
+++ b/kartograf/context.py
@@ -47,7 +47,6 @@ class Context:
 
         self.data_dir_irr = str(Path(self.data_dir) / "irr")
         self.data_dir_rpki_cache = str(Path(self.data_dir) / "rpki" / "cache")
-        self.data_dir_rpki_tals = str(Path(self.data_dir) / "rpki" / "tals")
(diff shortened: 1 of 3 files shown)


fjahr:
Even when 9.5 is released this requires some robust testing of course.

fjahr:
Closing as superseeded by #154

PR #71, opened 2025-03-28
ci: Don't run pylint twice

fjahr:
Somehow the current config means the pylint ci gets run twice (see for example #70), so change that by using the same style as in the other ci jobs.


jurraca:
ACK

PR #73, opened 2025-04-07
tests: parsing collectors initial test

jurraca:
Basic parsing test for Routeviews. Mimicks other parsing tests we have: 
- updates test context with `collectors` entities
- adds a fixtures file to `tests/data`
- tests that the output file exists, and that specific parsing occurs, asserting on expected output list.

(lmk if I should squash some of this)


tests/test_collectors_parse.py
```diff
@@ -0,0 +1,36 @@
...
+    parse_routeviews_pfx2as(context)
+
+    result_file = Path(context.out_dir_collectors) / "pfx2asn_clean.txt"
+
+    assert result_file.exists()
+
+    with open(result_file, 'r') as f:
+        results = [l.strip() for l in f.readlines()]
+
+    assert "127.0.0.0/24 AS12345" not in results
```
fjahr, line 32: nit: Could have added comments why these are not included in the result but can be done in a following test PR

fjahr:
ACK, looks great!

PR #76, opened 2025-04-24
update flake lock rpki-client-nix to correct url

jurraca:
meant to do this before we merged the upgrade to 9.5. I was testing with local paths so the `flake.lock` cached those. this updates to the correct remote (github repo) URL.


fjahr:
ACK

PR #77, opened 2025-04-24
Coverage fixes and test

jurraca:
A couple improvements to the `coverage` command.
It expects well-formed IP networks and addresses, so it will throw an error if you pass a malformed IP or an onion address.
Addtionally, the lookup for inclusion was considering the whole map file, when it can exit once it's found at least one valid inclusion. This can be further improved along the lines of what we did with `merge` but this is a start.
Finally, `02802d9e` adds a test for the coverage process with basic fixtures (valid IPv4, valid IPv6, onion addr that gets ignored).


kartograf/coverage.py
```diff
@@ -11,28 +11,34 @@ def coverage(map_file, ip_list_file):
...
 
     addrs = []
     for line in ip_list_file:
-        ip = int(ipaddress.ip_address(line.rstrip('\n')))
-        addrs.append(ip)
+        try:
+            ip = ipaddress.ip_address(line.rstrip('\n'))
+            addrs.append(int(ip))
+        except ValueError:
+            continue
```
jurraca, line 34: printing the invalid IPs here can get quite noisy if the user fetched those IPs from their node and there are onion or i2p addresses, so I opted to skip silently. 
alternatively, we could keep a tally of invalids (optionally by type).

fjahr: I would prefer it if we fail on the first invalid and tell the user the content of the offending line. This is easier and better than continuing and showing a result at the end that may not actually contain what the user is expecting because they provided the wrong data without knowing. If they provided onion addresses they must have misunderstood something and I think it's reasonable that they filter their list themselves. But they could also simply have a formatting issue in their IPv6s or something and then the result could show plausible but incorrect data.

tests/test_coverage.py
```diff
@@ -0,0 +1,18 @@
...
+def test_coverage(capsys):
+    map_file_path, ip_list_path = fixtures()
+    with open(map_file_path, 'r') as f:
+        map_items = f.readlines()
+    with open(ip_list_path, 'r') as f:
+        ip_items = f.readlines()
+
+    coverage(map_items, ip_items)
+    captured = capsys.readouterr()
+    assert "A total of 2 IPs out of 2 are covered by the map. That's 100.00%" in captured.out
```
fjahr, line 18: There should also be an IP in the list that is not covered.

kartograf/coverage.py
```diff
@@ -11,7 +11,11 @@ def coverage(map_file, ip_list_file):
     rpki_masks = []
     for line in map_file:
         pfx, _ = line.split()
-        ipn = ipaddress.ip_network(pfx)
+        try:
+            ipn = ipaddress.ip_network(pfx)
+        except ValueError:
+            print(f"Invalid IP network provided: {line}")
+            continue
```
fjahr, line 19: I think it would make sense to stop here as well just for the other file

kartograf/coverage.py
```diff
@@ -31,7 +32,11 @@ def coverage(map_file, ip_list_file):
             ip = ipaddress.ip_address(line.rstrip('\n'))
             addrs.append(int(ip))
         except ValueError:
-            continue
+            print(f"""
+                  Invalid IPv4/IPv6 address provided: {line}.
+                  Please remove and re-run.
+                  """)
+            sys.exit()
```
fjahr, line 39: Hm, I think I would have just raised an error with the message instead of doing it like this. Did you go this route to make a nicer output for the user?

kartograf/coverage.py
jurraca: I updated this to raise, but assuming this is from an ASmap file generated by kartograf, it's unlikely to occur since we already ensure validity of networks we write.

kartograf/coverage.py
jurraca: yea fair point, no particular reason I went this route. Updated.

kartograf/coverage.py
fjahr: > assuming this is from an ASmap file generated by kartograf, it's unlikely to occur since we already ensure validity of networks we write.

Thanks! Right, but if something is wrong here that's more reason to me to fail loudly and make the user aware that they likely have made a mistake rather than showing them a result that isn't of use to them.

fjahr:
ACK

PR #78, opened 2025-04-24
rpki: Sync cache up to three times in collaborative run

fjahr:
This implements the idea suggested here: https://github.com/asmap/kartograf/issues/69#issuecomment-2762459730 (Number 2 at the bottom).

In the case of a collaborative run (using the wait feature), kartograf will use the following behavior:
- There is a predefined timeframe for sync caching. I chose 10 minutes for this but it needs further testing if this needs to be expanded. It worked for me on slowed down connections well enough but I didn't test different CPU performance scenarios.
- On start, sync the cache once right away
- Then, check how much time is still left of the predefined timeframe. If there is still 30% of time left, just run again right away for a second time. If not, wait until the time is about to run out and run one more time at the end of the predefined period.
- After the second run has finished just wait until the time is running out and then run a third time.

Rationale:
- Even on a very fast computer with fast connection the sync will run 3 times at most, even if probably 5-6 time were possible. I have decided to do it this way because I am not sure syncing more than 3 times can give us much better results. If sending many requests to the smaller repositories is indeed a problem for them like we discussed, then this is a measure to not DDOS them too much. And it's also easier to reason about this since it's more constrained and we are still looking to learn from these results initially.
- I am hoping that most modern laptops with a stable home connection should be able to sync 3 times within 10 minutes. If not I would want to expand the windo to 15min.
- Slower hardware like a cheap VPS or a Raspberry PI should at least be able to run twice at least within the used timeframe.
- The logic is nicely contained in context, but happy to debate if it should be moved somewhere else. It's only used by rpki, so maybe it should rather be moved there.

Open questions:
- Should we expand the default timeframe as mentioned above (needs testing)
- I am still undecided if I want to add an option to change the timeframe. We want all people to be use the same timeframe so this could be a footgun but on the other hand it may make testing easier in the beginning.

This is what the output currently looks like:

```
Downloading RPKI Data, this may take a while.
RPKI sync #1
...took 180 seconds
RPKI sync #2
...took 191 seconds
(now sleeping 224 seconds)
RPKI sync #3
...took 142 seconds
Downloaded RPKI Data, hash sum: 821a68e84dc69ceb09c4753020c67d45567edee6b08533f50143e72605583bc9
...finished in 0:12:54.375893
```

(In this particular one the second took longer than the first, which is surprising. It was on a fast connection so download speed may just not be a blocker. But most tests took longer on the first sync.)

PR #80, opened 2025-04-29
IRR: Switch from FTP to https download

fjahr:
Fixes the IRR FTP download errors encountered with the ARIN DB here: https://github.com/asmap/kartograf/pull/78#issuecomment-2839095170

ARIN has dropped support for FTP so the easiest fix seems to be to switch IRR fetching over to https: https://www.arin.net/blog/2025/02/10/ftp-retirement/

### DIFF
--- a/kartograf/irr/fetch.py
+++ b/kartograf/irr/fetch.py
@@ -1,25 +1,25 @@
-from ftplib import FTP
 import gzip
 import shutil
 import time
 from pathlib import Path
+import requests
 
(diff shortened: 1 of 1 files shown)


jurraca:
tACK [1e95e6c](https://github.com/asmap/kartograf/pull/80/commits/1e95e6c724e67b6d1251804ce2d7f90d5d391b99)
makes sense to move them all the `https`.

PR #85, opened 2025-07-02
add analysis script of data sources

jurraca:
A script to analyze the data provenance for a given `out/_timestamp` folder. 

Reports raw and unique counts, overlap, proportion of our data sources.

Math seems ok, example result run:
<details><summary> example</summary>

```
=== RPKI, IRR, and Routeviews Data Analysis ===

Loading RPKI data...
RPKI networks loaded: 592452
Loading IRR data...
IRR networks loaded: 2097091
Loading Routeviews data...
Routeviews networks loaded: 1271102

============================================================

1. IRR data, relative to RPKI base:
--------------------------------------------------
Total IRR networks: 2,097,091
IRR networks already in RPKI: 379,682
Unique IRR networks: 1,717,409
Percentage of IRR data that is unique: 81.89%
* After merge *
Total networks after IRR merge: 1,068,290
Net networks added by IRR: 475,838
IRR increased dataset by: 80.32%

2. Routeviews data, relative to RPKI + IRR base:
--------------------------------------------------
Total Routeviews networks: 1,271,102
Routeviews networks already in RPKI+IRR: 528,721
Unique Routeviews networks (added to base): 742,381
Percentage of Routeviews data that is unique: 58.40%

3. Combined data (RPKI + IRR + Routeviews):
--------------------------------------------------
RPKI base networks: 592,452
Final combined networks: 1,412,296
Total networks added by IRR + Routeviews: 819,844
Combined sources increased dataset by: 138.38%

4. Cross-Source Analysis:
--------------------------------------------------
Networks common to IRR and Routeviews: 542,769
Networks unique to IRR (not in RPKI or Routeviews): 1,461,862
Networks unique to Routeviews (not in RPKI or IRR): 639,374
```

</details>

### DIFF
--- a/scripts/data_provenance_analysis.py
+++ b/scripts/data_provenance_analysis.py
@@ -1,0 +1,150 @@
+#!/usr/bin/env python3
+"""
+Analysis script to compare RPKI, IRR, and Routeviews data contributions.
+Shows how much additional IP network data each source adds to the final dataset.
+"""
+
+import ipaddress
+from pathlib import Path
+import sys
+from typing import Set, Dict
+import argparse
+
+def load_networks_from_file(file_path: Path) -> Set[str]:
+    """Load IP networks from a file, returning a set of prefixes."""
+    networks = set()
+    if not file_path.exists():
+        print(f"Warning: File {file_path} does not exist")
+        return networks
+
+    with open(file_path, 'r') as f:
+        for line in f:
+            line = line.strip()
+            if line and ' ' in line:
+                prefix = line.split()[0]
+                try:
+                    ipaddress.ip_network(prefix)
+                    networks.add(prefix)
+                except ValueError:
+                    print(f"Invalid prefix: {prefix}")
+                    continue
+    return networks
+
+def calculate_coverage_stats(base_networks: Set[str], additional_networks: Set[str]) -> Dict[str, int]:
+    """Calculate statistics about how much additional data a source provides."""
+    total_additional = len(additional_networks)
+    unique_to_additional = len(additional_networks - base_networks)
+    overlapping = len(additional_networks & base_networks)
+
+    return {
+        'total_additional': total_additional,
+        'unique_to_additional': unique_to_additional,
+        'overlapping': overlapping,
+        'percentage_unique': (unique_to_additional / total_additional * 100) if total_additional > 0 else 0
+    }
+
+def analyze_data_contributions(out_dir: Path) -> None:
+    """Analyze the data contributions from different sources."""
+
+    rpki_file = out_dir / "rpki" / "rpki_final.txt"
+    irr_file = out_dir / "irr" / "irr_final.txt"
+    routeviews_file = out_dir / "collectors" / "pfx2asn_clean.txt"
+
+    rpki_irr_merged = out_dir / "merged_file_rpki_irr.txt"
+    rpki_irr_rv_merged = out_dir / "merged_file_rpki_irr_rv.txt"
+
+    print("=== RPKI, IRR, and Routeviews Data Analysis ===\n")
+
+    # Load base RPKI data
+    print("Loading RPKI data...")
+    rpki_networks = load_networks_from_file(rpki_file)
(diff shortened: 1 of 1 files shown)


fjahr:
Cool, I think having some more analysis on this data will be very helpful! I didn't look at the code in detail yet, just a few comments on the results from my example run (my local data from `1749744000`):

```
=== RPKI, IRR, and Routeviews Data Analysis ===

Loading RPKI data...
RPKI networks loaded: 582986
Loading IRR data...
IRR networks loaded: 2221767
Loading Routeviews data...
Routeviews networks loaded: 1268693

============================================================

1. IRR data, relative to RPKI base:
--------------------------------------------------
Total IRR networks: 2,221,767
IRR networks already in RPKI: 376,531
Unique IRR networks: 1,845,236
Percentage of IRR data that is unique: 83.05%
* After merge *
Total networks after IRR merge: 1,163,583
Net networks added by IRR: 580,597
IRR increased dataset by: 99.59%

2. Routeviews data, relative to RPKI + IRR base:
--------------------------------------------------
Total Routeviews networks: 1,268,693
Routeviews networks already in RPKI+IRR: 528,284
Unique Routeviews networks (added to base): 740,409
Percentage of Routeviews data that is unique: 58.36%

3. Combined data (RPKI + IRR + Routeviews):
--------------------------------------------------
RPKI base networks: 582,986
Final combined networks: 1,503,498
Total networks added by IRR + Routeviews: 920,512
Combined sources increased dataset by: 157.90%

4. Cross-Source Analysis:
--------------------------------------------------
Networks common to IRR and Routeviews: 541,394
Networks unique to IRR (not in RPKI or Routeviews): 1,590,387
Networks unique to Routeviews (not in RPKI or IRR): 636,831
```

I'm not sure the "unique networks" is the most helpful metric. For example in 1: `Unique IRR networks: 1,845,236` but then `Net networks added by IRR: 580,597` and also `RPKI base networks: 582,986`, this seems weird but can probably be explained by the fact that the RPKI network entries on average cover more IP space than the IRR networks. But this would need to be analyzed further and if that is necessary then the analysis here doesn't provide as much value as it could. I think it's better to look at the numbers in terms of IP space coverage % rather than unqiue networks. I used to run some analysis like that in the past and I think for this kind of analysis it's better because it's more intuitive. In the logging of the runs we use the networks because it's good ok to compare between different users runs if one of them had problems with the sync and missed some repo etc. 

`Networks unique to IRR (not in RPKI or Routeviews): 1,590,387` here this also shows how the "uniqueness" view is problematic IMO. It looks like IRR provides a large amount of IP space that the others sources are not aware of but I think that's not really true. I think it's at least in part that RPKI has something like X.X.X.X/16 -> AS123 and IRR has a bunch of variations of X.X.X.X/24 -> AS123. The view of coverage of the IP space would show this more intuitively: relative to the RPKI network the IRR networks offer less coverage in isolation and no additional coverage. Calling them unique sounds too much like they are valuable when they are actually not. But it would be good to report if there were some subnets that didn't map to AS123. We can keep the network numbers around as well of course, maybe we have to distinguish between "unique mapping" and "unique coverage" or something like that. And if we are speaking of overlap I would expect that the case described above identifies these networks as overlapping (because I am thinking of coverage) while I think currently they would not be reported as overlapping. 

I am also thinking if we may just want to print out these numbers as part of our normal run logs and potentially replace some of the network numbers with the coverage numbers. Leave some stuff out but report after every merge how much the source has contributed and how much was discarded and of the discarded how much was overlaps with the same ASN and how much was conflicting.

Keeping it as a separate script seems fine too though. The next step that would be good to add here is to have a list of IPs (nodes) and show which sources covered how many of the IPs in the list. That's of course something that only makes sense to do in a script.

Another thought on the IP space, we could report relative to total IP space but I think that's annoying especially for IPv6. It would make more sense to load the final result file first, take the total coverage of IP space out of that and then report the relative numbers to that from each of the sources.

I also think this script can be simpler if you reuse some of our internal functions.

jurraca:
closing. good feedback, and I'd like to approach this data analysis story differently, i.e. not throwing a bunch of stats to the user like this.

PR #86, opened 2025-07-13
Cleanup out dir intermediate files, set debug to False

jurraca:
Deletes intermediate files created by the kartograf run by default.

The `-d/--debug` flag is set to `False` by default. Setting it to `True` prevents the file deletions, and preserves intermediate files on disk.

Note that this is a different process from the `--cleanup` flag, which deletes the `data/` directory input files. The intermediate files concerned here exist in the `out/` directory. To delete both input files and intermediate output files, the user can use the `-c` cleanup flag alone. To keep everything, just `-d` will work. This is a bit awkward in that deletion is by default for the `out` directory but opt-in for the `data` directory. We can rectify this in a follow-up if appropriate.


fjahr:
ACK, tested this several times with a different configurations.

PR #87, opened 2025-07-13
Optimize merge with chunked processing

jurraca:
A Claude suggestion: by chunking, we remove the need to build up a large list in memory, resulting in faster processing.

Tested this and merging does seem a fair bit faster.


fjahr:
I don't tink I understand how this changes the memory footprint, `extra_included` is still built and held in memory, isn't it?

Did you see a large improvement? When I just ran a full map with and without the change it was actually a bit slower with the change here but only slightly.  Did you use something specific to benchmark this?

PR #89, opened 2025-08-14
test_cli: remove single-quoted cli arguments

jurraca:
bug reported [here](https://github.com/asmap/kartograf/issues/88)
see https://github.com/python/cpython/pull/130751
this is a bug in cpython. The argument parsing behavior was changed in between two minor releases of Python 12. In order to continue working across Python versions, we relax the error message check.

Note that I was unable to reproduce this via Nixpkgs Python versions 3.11-3.13. Please test with Python downloaded from PIP.

### DIFF
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -83,7 +83,7 @@ def test_invalid_command(parser, capsys):
     with pytest.raises(SystemExit):
         parser.parse_args(['invalid'])
     captured = capsys.readouterr()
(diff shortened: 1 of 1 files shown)


fjahr:
tACK, fixes the failing test in the setup where I could reproduce the test breaking previously.

@jurraca just FYI, if you include the following in your PR description then github links the issue to the PR and automatically closes the issue when the PR is merged:

Closes #88

PR #91, opened 2025-09-19
Print logs cleanup

jurraca:
We currently have some fairly verbose log printing, which was set up to be extra-helpful when getting the script functionality into place. With more stability, and the defaulting of `debug` to False, it's reasonable to revisit some of the printed logs. This also includes a small fix to logs in `debug.log`.

These changes remove, compact, or re-format some printed logs, and do not affect any input or output functionality.

The largest removal is the "extra filtered file" statements from the merge process. This is an internal process to the function so the log is both confusing and unnecessary. Also, we update the logs to the `debug.log` file to not print prefixes if they are not parsed as such. Currently, the behavior is to print `parser encountered an invalid IP network: None`.

### DIFF
--- a/kartograf/collectors/parse.py
+++ b/kartograf/collectors/parse.py
@@ -60,7 +60,9 @@ def parse_routeviews_pfx2as(context):
             prefix = parse_pfx(prefix)
             asn = asn.upper().rstrip('\n')
 
-            if not prefix or is_bogon_pfx(prefix) or is_bogon_asn(asn):
+            if not prefix:
+                continue
+            if is_bogon_pfx(prefix) or is_bogon_asn(asn):
                 if context.debug_log:
                     with open(context.debug_log, 'a') as logs:
                         logs.write(f"Routeviews: parser encountered an invalid IP network: {prefix}")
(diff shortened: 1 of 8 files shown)


kartograf/collectors/parse.py
```diff
@@ -60,7 +60,9 @@ def parse_routeviews_pfx2as(context):
             prefix = parse_pfx(prefix)
             asn = asn.upper().rstrip('\n')
 
-            if not prefix or is_bogon_pfx(prefix) or is_bogon_asn(asn):
+            if not prefix:
+                continue
```
fjahr, line 63: Hm, unsure if we shouldn't print something to debug in this case since there might be some issue that we should fix. Something like "couldn't parse a prefix from {line}" or so?

fjahr: (And of course the other prefix parse too)

fjahr (commented):
Looks mostly good, I will test once we have agreed on something about the logging of the "no prefix" case. Also the pylint CI complains about something but that seems more like a case for an ignore than something else.

jurraca:
Updated with your suggestions. I kept the Routeviews URL, since it would be hard to guess where to go look for the source, so imo its nice to have it in the logs as a full URL.

jurraca:
i'll squash this before merging yea?

PR #95, opened 2025-10-01
download IRR databases concurrently

### LINKED ISSUE #94: Concurrent downloads for IRR DBs
Another idea I had that could squeeze out a little additional performance: We could try downloading the IRR files concurrently. Routeviews is just two files and much smaller and hosted on the same server, so not relevant and RPKI sync is whole different story.

Needs to be tested of course how big the impact is and there should be no stability issues. If network connection is a bottleneck I guess this should not have a negative impact I hope, but should probably be also challenged.

jurraca:
Downloading IRR DBs concurrently seems to be a bit faster and is generally a reasonable thing, though as with all parallelism, we're bottlenecked by our slowest item, which in this case is almost always AFRINIC.  

closes #94

### DIFF
--- a/kartograf/irr/fetch.py
+++ b/kartograf/irr/fetch.py
@@ -1,3 +1,4 @@
+from concurrent.futures import ThreadPoolExecutor, as_completed
 import gzip
 import shutil
 import time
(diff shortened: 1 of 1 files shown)


fjahr:
Cool, will try this out soon.

> we're bottlenecked by our slowest item, which in this case is almost always AFRINIC

Right, I wasn't expecting a huge jump from this but having big differences in the download speeds is actually the scenario where this makes the most sense, so the one slow download isn't blocking the others 👍

fjahr:
Works great for me, code looks good. Afrinic IRR download seems to be extremely slow at the moment but I confirmed that this is unrelated to this change.

fjahr:
ACK

PR #97, opened 2025-10-21
use -p threads flag

jurraca:
Use the `-p` threads flag from `rpki-client` 9.6.

This is a draft of a PR that is a breaking change, since previous versions of `rpki-client` do not have this flag, and this should not be merged until we have a strategy for that.

The changes simplify the code, but we lose information about `valid_until` and `valid_since` which was previously used to select the most likely useful announcement. Thus our "smallest ASN wins" heuristic is doing work than previously.

On average I get a runtime for parsing of about 45 seconds compared to 3-4 minutes previously. 

@fjahr

### DIFF
--- a/kartograf/merge.py
+++ b/kartograf/merge.py
@@ -2,13 +2,12 @@ from concurrent.futures import ProcessPoolExecutor
 from pathlib import Path
 import ipaddress
 import math
-import os
 import shutil
 from types import SimpleNamespace
 import pandas as pd
 
 from kartograf.timed import timed
-from kartograf.util import get_root_network
+from kartograf.util import get_root_network, get_threads
 
 
 class BaseNetworkIndex:
@@ -163,7 +162,7 @@ def pick_chunk_size(n_rows: int, workers: int | None = None,
                     min_chunk: int = 5,
                     max_chunk: int = 200_000) -> int:
     if workers is None:
-        workers = os.cpu_count() or 4
+        workers = get_threads()
     chunk = math.ceil(n_rows / workers)
     return max(min_chunk, min(max_chunk, chunk))
 


kartograf/rpki/fetch.py
```diff
@@ -102,6 +103,9 @@ def validate_rpki_db(context):
 
     tal_options = [item for path in data_tals(context) for item in ('-t', path)]
 
+    cpu_count = os.cpu_count()
+    threads = cpu_count if cpu_count else 4
```
fjahr: Since [#98](https://github.com/asmap/kartograf/pull/98) is merged now, please put this threads count stuff into some util function that is shared between the two different places.

jurraca:
Looking at the CI reproduction failure, this is what I was somewhat concerned about using `expiry` instead of `valid_since` and `valid_until` checks. 
The diff between running the CI input data with this branch and master is 22185, of which 6 are definitely not due to the "lower ASN wins" heuristic. But the rest were originally (master) assigned differently due to the `valid_until` or `valid_since` heuristic, or the "lower ASN wins" heuristic. 

(the six outliers, here the "older" file is the master branch file)
```
27.124.87.0/24 AS41239 # was AS9341
27.124.88.0/24 AS41239 # was AS9341
103.162.141.0/24 AS141668 # was AS141083
103.166.234.0/24 AS141963 # was AS140407
119.2.64.0/19 AS131704 # was AS17450
202.47.66.0/24 AS150228 # was AS138106
```

All the ROAs are valid and so it's not that we're issuing bad ROAs, but this big of a change between two runs is difficult to justify, especially since our current heuristics are more precise than the ones in this branch.

PR #98, opened 2025-10-29
use ProcessPoolExecutor to merge files

jurraca:
copy the base_dict to separate OS processes and run the `contains_row` check in parallel. This speeds up merge by ~2.5x on my machine.

I tested against the previous implementation and got the same hash, showing that we're not changing the logic and outcome. Curious how this fares on a Mac.

### DIFF
--- a/kartograf/merge.py
+++ b/kartograf/merge.py
@@ -188,22 +188,10 @@ def general_merge(
     chunks = []
     chunk_data = []
     for i, row in df_extra.iterrows():
-        if i == 0:
-            chunk_data.append((i, (row.INETS, row.PFXS, row.PFXS_LEADING)))
-            continue
-        # reached the chunk size
-        if i > 0 and ((i % chunk_size) == 0):
-            chunk_data.append((i, (row.INETS, row.PFXS, row.PFXS_LEADING)))
+        chunk_data.append((i, (row.INETS, row.PFXS, row.PFXS_LEADING)))
+        if (i + 1) % chunk_size == 0 or i == len_df_extra - 1:
             chunks.append(chunk_data)
-        # reached the final row
-        elif i == (len_df_extra - 1):
-            chunk_data.append((i, (row.INETS, row.PFXS, row.PFXS_LEADING)))
-            chunks.append(chunk_data)
-        else:
-            chunk_data.append((i, (row.INETS, row.PFXS, row.PFXS_LEADING)))
-            continue
-        # reset chunk data
-        chunk_data = []
+            chunk_data = []
 
     all_results = []
     with ProcessPoolExecutor() as executor:


fjahr (commented):
Awesome! I did one reproduction and I did indeed see a similar speed-up and the same result.

The linter is complaining and I found the code a bit hard to follow on the first pass. Could you try to make it a bit easier by adding some comments maybe and looking if it can be simplified a little more?

kartograf/merge.py
```diff
@@ -139,20 +164,37 @@ def general_merge(
...
-            base.update(pfx)
+            base_network_index.update(pfx)
 
     df_extra = extra_file_to_df(extra_file)
 
-    extra_included = []
-    for row in df_extra.itertuples(index=False):
-        result = base.contains_row(row)
-        extra_included.append(result)
+    chunk_size = 50000
```
fjahr: The optimal chunk size will probably depend on the machine, can you try if you can get similar or better results with dynamic picking of the chunk size with something like this (untested):

```python
def pick_chunk_size(n_rows: int, workers: int | None = None,
                    tasks_per_worker: int = 16,
                    min_chunk: int = 10_000,
                    max_chunk: int = 200_000) -> int:
    if workers is None:
        workers = os.cpu_count() or 4
    num_tasks = max(1, workers * tasks_per_worker)
    chunk = math.ceil(n_rows / num_tasks)
    return max(min_chunk, min(max_chunk, chunk))
```

kartograf/merge.py
```diff
@@ -139,20 +164,37 @@ def general_merge(
...
+        chunk = df_extra.iloc[i:i+chunk_size]
+        # get each chunk with an index
+        chunk_data = [
+            (i+j, (row.INETS, row.ASNS, row.PFXS, row.PFXS_LEADING))
+            for j, row in enumerate(chunk.itertuples(index=False))
+        ]
+        chunks.append(chunk_data)
+
+    with ProcessPoolExecutor() as executor:
+        base_dict = base_network_index.get_serializable_dict()
```
fjahr: Is there a reason why the `base_dict` is handed around and not and instance of `BaseNetworkIndex`? I think it would be cleaner unless there there is some reason that prevents that.

jurraca: yea, the _dict.keys() thing is not picklable, and everything passed to ProcessPollExecutor must be picklable.

kartograf/merge.py
```diff
@@ -171,8 +213,8 @@ def general_merge(
             None, sep=" ", index=False, columns=["PFXS", "ASNS"], header=False
         )
 
-    with open(base_file, "r") as base:
-        base_contents = base.read()
+    with open(base_file, "r") as f:
```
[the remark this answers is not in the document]
jurraca: yep, rolled these renaming changes back.

jurraca:
> Some tests failed in the CI after the latest push, in case you missed it.

took me a bit to rework it :) should be ready for a re-review.

kartograf/merge.py
```diff
@@ -139,20 +175,49 @@ def general_merge(
...
+    len_df_extra = len(df_extra)
+    chunk_size = pick_chunk_size(len_df_extra)
+    chunks = []
+    chunk_data = []
+    for i, row in df_extra.iterrows():
+        if i == 0:
+            chunk_data.append((i, (row.INETS, row.PFXS, row.PFXS_LEADING)))
+            continue
+        # reached the chunk size
+        if i > 0 and ((i % chunk_size) == 0):
```
fjahr: Shouldn't this be:

```suggestion
        if i > 0 and ((i+1 % chunk_size) == 0):
```

kartograf/merge.py
```diff
@@ -131,6 +147,26 @@ def extra_file_to_df(extra_file_path):
...
+        result = base.contains_row(row)
+        results.append((original_idx, result))
+    return results
+
+
+def pick_chunk_size(n_rows: int, workers: int | None = None,
+                    min_chunk: int = 5,
+                    max_chunk: int = 200_000) -> int:
+    if workers is None:
+        workers = (os.cpu_count() - 2) or 4
```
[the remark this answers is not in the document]
jurraca: yea, was trying to leave some CPUs free but it shouldn't matter so i reverted it.

kartograf/merge.py
```diff
@@ -131,6 +147,26 @@ def extra_file_to_df(extra_file_path):
 
     return df_extra
 
+def process_chunk_worker(chunk_data, base_dict):
+    base = BaseNetworkIndex.from_dict(base_dict)
```
fjahr: On my latest pass I have noticed that we are pickling this per chunk instead of per worker. There should be a way to pickle this per worker and keep it for as many chunks as the worker is processing. But we can keep this as a follow-up optimization, maybe along with the tasks per worker idea. The performance impact would need to be measured across different setups to check if there is any serious impact.

tests/test_merge.py
```diff
@@ -122,3 +122,14 @@ def test_merge_joint(tmp_path):
...
+def test_pick_chunk_size():
+    '''
+    Test picking a chunk size for the merge function.
+    '''
+    assert pick_chunk_size(100, workers=16) == 7
+    assert pick_chunk_size(100, workers=4) == 25
+    # min_chunk wins
+    assert pick_chunk_size(10, workers=4) == 5
+    # min_chunk wins
+    assert pick_chunk_size(0) == 5
```
fjahr: nit: A max_chunk case could have been added as well.

PR #100, opened 2025-11-04
cli: Make it possible to launch with `python -m kartograf.cli`

laanwj:
Make  it possible to launch the cli with `python -m kartograf.cli`. This would avoid having to rename the `run` wrapper in #99.


fjahr:
tACK 48f3d8dab15f1eb7d27662ec58ebade09ad76f50

Makes sense and shouldn't come with any downsides. Thanks!

PR #104, opened 2025-11-07
bug: bring back reproduction run CI job

jurraca:
the actions runner doesn't like output with linebreaks, so we replace all newlines with spaces in the output, and check for the presence of the hash in that string.


fjahr:
ACK

Merging this now so we are back to the old state that we should have had, debugging the other problem and updating the test to catch it can come later.

PR #110, opened 2025-12-04
fix: run python3 with line buffering on the `run` script

luisschwab:
Python uses full buffering when `stdout` is not a `tty` (eg when piping to `tee`), which prevents real-time output visibility. This PR adds the `-u` flag to enable line buffering, allowing output to be displayed immediately after each newline.

This ensures that commands like `./run map -w=1764864000 -irr -rv -s | tee run.txt` display output in real-time rather than buffering until completion.

### DIFF
--- a/run
+++ b/run
@@ -1,4 +1,4 @@
-#!/usr/bin/env python3
+#!/usr/bin/env -S python3 -u
 """
 This project allows to build a IP prefix to AS map using RPKI as the primary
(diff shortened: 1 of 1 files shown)


jurraca:
utACK https://github.com/asmap/kartograf/pull/110/commits/35466cd4da27c5f4e5f202e3296e3831f6d54008

thanks! this is a nice fix :+1:

PR #116, opened 2026-01-09
Use trie in cov command

jurraca:
rebased on #114 and #115

Uses the trie implementation from #115 in the `cov` command.

This changes the functionality in one important way: previously, the lookup returned the first network it found which matched the IP address. But if there was a more specific prefix (i.e. a more accurate home for it) later in the list, it would not reach it. The trie lookup always finds the longest matching prefix, which aligns with internet routing lookups.

### DIFF
--- a/kartograf/coverage.py
+++ b/kartograf/coverage.py
@@ -1,82 +1,62 @@
 import ipaddress
-import numpy as np
-import pandas as pd
+from kartograf.trie import IPTrie
 
 
 def coverage(map_file, ip_list_file, output_covered=None, output_uncovered=None):
-    nets = []
-    masks = []
-    asns = []
-    for line in map_file:
-        pfx, asn = line.split()
-        try:
-            ipn = ipaddress.ip_network(pfx)
-        except ValueError:
-            raise ValueError(f"""
-                  Invalid IP network provided: {line}
-                  Please remove and re-run.
-                  """)
+    print("Running coverage check...\n")
 
-        netw = int(ipn.network_address)
-        mask = int(ipn.netmask)
-        masks.append(mask)
-        nets.append(netw)
-        asns.append(asn)
-
-    net_masks = np.array(masks)
-    network_addresses = np.array(nets)
-    zipped = list(zip(net_masks, network_addresses, asns))
-
+    trie = IPTrie()
+    trie.from_map_file(map_file)
     addrs = []
     for line in ip_list_file:
+        line = line.rstrip('\n')
+        if not line:
+            continue
         try:
-            ip = ipaddress.ip_address(line.rstrip('\n'))
-            addrs.append(int(ip))
+            ip = ipaddress.ip_address(line)
+            addrs.append((ip, line))
         except ValueError:
             raise ValueError(f"""
                   Invalid IPv4/IPv6 address provided: {line}.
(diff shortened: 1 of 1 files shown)


fjahr:
> This changes the functionality in one important way: previously, the lookup returned the first network it found which matched the IP address. But if there was a more specific prefix (i.e. a more accurate home for it) later in the list, it would not reach it. The trie lookup always finds the longest matching prefix, which aligns with internet routing lookups.

I'm not sure I could call this is a change in behavior or maybe I am missing something. For the coverage statistics it doesn't matter which ASN covers it, right? But it probably changes the output parameters, but these aren't merged yet.  I would actually call that a bug of the output functionality if I am understanding this correctly and maybe we shouldn't merge it like this then. Maybe the trie implementation should go in first and then these cov output changes is added with using the trie right away. Just seems cleaner :)

jurraca:
rebased and ready for review @fjahr

kartograf/coverage.py
```diff
@@ -1,7 +1,8 @@
 import ipaddress
 from kartograf.trie import IPTrie
+from kartograf.timed import timed
 
-
+@timed
```
fjahr, line 5: I would prefer if we get rid of this again. Since the command is very fast now there is no need for it IMO.

kartograf/coverage.py
```diff
@@ -1,82 +1,62 @@
...
-        netw = int(ipn.network_address)
-        mask = int(ipn.netmask)
-        masks.append(mask)
-        nets.append(netw)
-        asns.append(asn)
-
-    net_masks = np.array(masks)
-    network_addresses = np.array(nets)
-    zipped = list(zip(net_masks, network_addresses, asns))
+    print("Running coverage check...\n")
```
fjahr, line 7: I don't think we need this print here, just returning the result as we did before would be more clean IMO. The newline also isn't necessary and makes the output look a bit weird for my taste.

```
Running coverage check...

A total of 7545 IPs out of 10000 are covered by the map. That's 75.45%
...finished in 0:00:06.320278
```

kartograf/coverage.py
```diff
@@ -1,82 +1,62 @@
...
+    for ip, raw_addr in addrs:
+        asn = trie.lookup(ip)
+        if asn is not None:
+            covered_pairs.append((raw_addr, asn))
+        else:
+            not_covered_ips.append(ip)
 
-    len_covered = len(df_cov)
-    total = len(df)
+    len_covered = len(covered_pairs)
```
fjahr, line 34: I have realized that if there are no IPs parsed then we are dividing by zero below and crash. This isn't new but should be fixed in a follow-up. I confirmed that this can happen by giving an empty file as the ip list. In that case we should probably just return early with appropriate feedback.

fjahr (commented):
tACK bdbd7eb1fe4f477b2b229f5e6d912f2074aff54d

I have tested this across the last two recent runs with three randomly generated IP lists of 10k IPs each as well as two recent dumps of Bitcoin network node IPs. The cov command returned the same result as master each time.

It would be great if the output could go back to how it was previously since I don't see a need to change it especially with the speedup we got now. But this can be done in a follow-up together with the existing divide by zero issue I found since this has been already open for such a long time.

PR #123, opened 2026-02-11
Use trie in merge process

jurraca:
Use the trie implementation in the merge process, replacing the concurrent, chunked, map-based process.

It turns out there was a bug in our implementation. When we called `contains_row` to check whether a network was included in the base, it only checks whether that candidate / incoming network is a subnet. But it could also be a super-net, i.e. a shorter prefix. In that case both would exist in the merged output, but would be logically conflicting; we want to consider the base (RPKI) entry as the valid one, and reject the candidate. 5b0eea38ec651b4bbc09a7bb67211c629335d92e ensures we never overwrite base/RPKI entries, whether the candidate is a subnet or supernet of an existing entry. 

Running a reproduction run against the latest collab run, the new final_result file is 3254 entries longer. A diff of the original final_result and the new one gives 15699 entries (1% of total), 95% (15003) of which were previously unassigned. This sounds about right, because we keep the RPKI entries which are often most specific. I'm not sure though.
The count of entries and file size stay about the same, as does the encoded (filled or unfilled) file size.

Minimal test changes, probably should add some edge case testing, but the merge tests pass without changes. 

Overall, this makes the merge code simpler and removes some dependencies. We can probably remove pandas entirely next with a few more changes. It also speeds up the merge process.

However this will break our reproducibility CI check.

### DIFF
--- a/kartograf/merge.py
+++ b/kartograf/merge.py
@@ -1,15 +1,11 @@
-from concurrent.futures import ProcessPoolExecutor
 from pathlib import Path
 import ipaddress
-import math
-import os
 import shutil
-from types import SimpleNamespace
 import pandas as pd
 
 from kartograf.timed import timed
+from kartograf.trie import IPTrie
 from kartograf.util import get_root_network
-
 
 class BaseNetworkIndex:
     '''
@@ -21,63 +17,33 @@ class BaseNetworkIndex:
     instead of all the networks in the base file.
     '''
 
-
     def __init__(self):
-        self._dict = {4: {}, 6: {}}
-        self._v4_keys = self._dict[4].keys()
-        self._v6_keys = self._dict[6].keys()
+        self._trie = IPTrie()
 
     def update(self, pfx):
         try:
-            ipn = ipaddress.ip_network(pfx)
+            ipn = ipaddress.ip_network(pfx, strict=False)
         except ValueError:
             print(f"Invalid prefix provided: {pfx}")
             return
-
-        netw = int(ipn.network_address)
-        mask = int(ipn.netmask)
-        v = ipn.version
-        root_net = get_root_network(pfx)
-
-        if (root_net in self._v4_keys) or (root_net in self._v6_keys):
-            current = self._dict[v][root_net]
-            self._dict[v][root_net] = current + [(netw, mask, ipn.prefixlen)]
-        else:
-            self._dict[v].update({root_net: [(netw, mask, ipn.prefixlen)]})
-
-    def check_inclusion(self, row, root_net, version):
-        """
-        A network is a subnet of another if the bitwise AND of its IP and the base network's netmask
-        is equal to the base network IP, and the network's prefix length is larger or equal than the base network's
-        prefix length.
-        """
-        candidate = ipaddress.ip_network(row.PFXS)
-        for net, mask, prefixlen in self._dict[version][root_net]:
-            if candidate.prefixlen >= prefixlen and row.INETS & mask == net:
-                return 1
-        return 0
+        self._trie.insert(pfx, ipn.prefixlen)
 
     def contains_row(self, row):
-        root_net = row.PFXS_LEADING
-        version = ipaddress.ip_network(row.PFXS).version
-        if version == 4 and (root_net in self._v4_keys):
-            return self.check_inclusion(row, root_net, version)
-        if version == 6 and (root_net in self._v6_keys):
-            return self.check_inclusion(row, root_net, version)
+        """
+        Check if the prefix in the row is covered by any prefix in the base file.
+        A candidate prefix is covered if:
+        1. Its network address matches a base prefix in the trie
+        2. The matching base prefix length <= candidate prefix length
+           (i.e., base prefix is equal or broader than candidate)
+        """
+        try:
+            candidate = ipaddress.ip_network(row.PFXS, strict=False)
+        except ValueError:
+            return 0
+        base_prefixlen = self._trie.lookup(candidate.network_address)
+        if base_prefixlen is not None and base_prefixlen <= candidate.prefixlen:
+            return 1
         return 0
-
-    def get_serializable_dict(self):
-        """Return the internal dict for serialization to worker processes."""
-        return self._dict
-
-    @classmethod
-    def from_dict(cls, data_dict):
-        """Reconstruct a BaseNetworkIndex from a serialized dict."""
-        instance = cls()
-        instance._dict = data_dict
-        instance._v4_keys = instance._dict[4].keys()
-        instance._v6_keys = instance._dict[6].keys()
-        return instance
 
 @timed
 def merge_irr(context):
@@ -149,34 +115,13 @@ def extra_file_to_df(extra_file_path):
 
     return df_extra
 
-def process_chunk_worker(chunk_data, base_dict):
-    base = BaseNetworkIndex.from_dict(base_dict)
-
-    results = []
-    for original_idx, (net_int, pfx, pfx_leading) in chunk_data:
-        row = SimpleNamespace(PFXS=pfx, PFXS_LEADING=pfx_leading, INETS=net_int)
-
-        result = base.contains_row(row)
-        results.append((original_idx, result))
-    return results
-
-
-def pick_chunk_size(n_rows: int, workers: int | None = None,
-                    min_chunk: int = 5,
-                    max_chunk: int = 200_000) -> int:
-    if workers is None:
-        workers = os.cpu_count() or 4
-    chunk = math.ceil(n_rows / workers)
-    return max(min_chunk, min(max_chunk, chunk))
-
-
 def general_merge(
     base_file, extra_file, extra_filtered_file, out_file
 ):
     """
     Merge lists of IP networks into a base file.
     """
-    print("Merging extra prefixes that were not included in the base file.")
+    print("Creating network index from base file.")
     base_network_index = BaseNetworkIndex()
     with open(base_file, "r") as file:
         for line in file:
@@ -185,28 +130,12 @@ def general_merge(
 
     df_extra = extra_file_to_df(extra_file)
 
-    len_df_extra = len(df_extra)
-    chunk_size = pick_chunk_size(len_df_extra)
-    chunks = []
-    chunk_data = []
-    for i, row in df_extra.iterrows():
-        chunk_data.append((i, (row.INETS, row.PFXS, row.PFXS_LEADING)))
-        if (i + 1) % chunk_size == 0 or i == len_df_extra - 1:
-            chunks.append(chunk_data)
-            chunk_data = []
+    print("Merging extra prefixes that were not included in the base file.")
+    extra_included = []
+    for row in df_extra.itertuples(index=False):
+        extra_included.append(base_network_index.contains_row(row))
 
-    all_results = []
-    with ProcessPoolExecutor() as executor:
-        base_dict = base_network_index.get_serializable_dict()
-        futures = [executor.submit(process_chunk_worker, chunk, base_dict) for chunk in chunks]
-
-        for future in futures:
-            all_results.extend(future.result())
-
-    # Sort by original index
-    all_results.sort(key=lambda x: x[0])
-
-    df_extra["INCLUDED"] = [result for _, result in all_results]
+    df_extra["INCLUDED"] = extra_included
 
     df_filtered = df_extra[df_extra.INCLUDED == 0]
 
--- a/kartograf/merge.py
+++ b/kartograf/merge.py
@@ -31,8 +31,7 @@ class BaseNetworkIndex:
     def contains_row(self, row):
         """
         Check if the prefix in the row is covered by any prefix in the base file.
-        A candidate prefix is covered if:
-        Its network address matches a base prefix in the trie
+        A candidate prefix is covered if its network address matches a prefix in the trie
         """
         try:
             candidate = ipaddress.ip_network(row.PFXS, strict=False)
--- a/kartograf/merge.py
+++ b/kartograf/merge.py
@@ -20,28 +20,26 @@ class BaseNetworkIndex:
     def __init__(self):
         self._trie = IPTrie()
 
-    def update(self, pfx):
+    def update(self, pfx, asn):
         try:
             ipn = ipaddress.ip_network(pfx, strict=False)
         except ValueError:
             print(f"Invalid prefix provided: {pfx}")
             return
-        self._trie.insert(pfx, ipn.prefixlen)
+        self._trie.insert(ipn, asn)
 
     def contains_row(self, row):
         """
         Check if the prefix in the row is covered by any prefix in the base file.
         A candidate prefix is covered if:
-        1. Its network address matches a base prefix in the trie
-        2. The matching base prefix length <= candidate prefix length
-           (i.e., base prefix is equal or broader than candidate)
+        Its network address matches a base prefix in the trie
         """
         try:
             candidate = ipaddress.ip_network(row.PFXS, strict=False)
         except ValueError:
             return 0
-        base_prefixlen = self._trie.lookup(candidate.network_address)
-        if base_prefixlen is not None and base_prefixlen <= candidate.prefixlen:
+        asn = self._trie.lookup(candidate)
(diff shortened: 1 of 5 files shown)


kartograf/merge.py
```diff
@@ -147,34 +115,13 @@ def extra_file_to_df(extra_file_path):
...
-
-
 def general_merge(
     base_file, extra_file, extra_filtered_file, out_file
 ):
     """
     Merge lists of IP networks into a base file.
     """
-    print("Merging extra prefixes that were not included in the base file.")
+    print("Creating network index from base file.")
```
fjahr, line 121: Why add this additional logging? I think we just recently removed similar lines to that extend, so I think keeping the logging as is would be better.

kartograf/trie.py
```diff
@@ -44,9 +44,27 @@ def insert(self, network, asn):
...
+        if isinstance(ip, (ipaddress.IPv4Address, ipaddress.IPv6Address)):
+            return self._lookup_address(ip)
+        raise TypeError("lookup expects an ip_address or ip_network object")
+
+    def _check_subtree(self, node):
+        '''
+        Check if subtree contains any networks recursively.
+        '''
+        if node.asn is not None:
+            result = node.asn
```
fjahr: A result is set here but if there are no children, then `None` is returned. Either this is a bug or I have not fully understood this code yet :)

kartograf/trie.py
```diff
@@ -71,6 +89,48 @@ def lookup(self, ip):
...
+
+        addr_int = int(network.network_address)
+        prefix_len = network.prefixlen
+
+        node = root
+        last_asn = None
+
+        # Traverse the trie to find RPKI networks that contain this candidate network
+        for i in range(max_bits):
+            if i<= prefix_len:
```
fjahr: missing space but also I think this needs to be `if i < prefix_len:` here to match the `range(prefix_len)` in the `insert` method. Seems like this currently takes one step too far.

fjahr (commented):
> It turns out there was a bug in our implementation. When we called contains_row to check whether a network was included in the base, it only checks whether that candidate / incoming network is a subnet. But it could also be a super-net, i.e. a shorter prefix. In that case both would exist in the merged output, but would be logically conflicting; we want to consider the base (RPKI) entry as the valid one, and reject the candidate. https://github.com/asmap/kartograf/commit/5b0eea38ec651b4bbc09a7bb67211c629335d92e ensures we never overwrite base/RPKI entries, whether the candidate is a subnet or supernet of an existing entry.

If I understand correctly what you are saying I don't think this is a bug. When we have overlapping coverage in our resulting asmap we use the most precise (longest) prefix for the ASN. If there is an overlapping, shorter entry we only use it for the additional coverage that it gets us aside from the one we got from the longer one. So these supernets never "overwrite" the RPKI coverage, they only add coverage where RPKI didn't apply. But maybe there is a misunderstanding about the issue on my side and an example might help, I didn't dig deeper into the code and resulting files on this yet.

> Running a reproduction run against the latest collab run, the new final_result file is 3254 entries longer. A diff of the original final_result and the new one gives 15699 entries (1% of total), 95% (15003) of which were previously unassigned. This sounds about right, because we keep the RPKI entries which are often most specific. I'm not sure though.
> The count of entries and file size stay about the same, as does the encoded (filled or unfilled) file size.

This result seems counterintuitive since there are more entries even though you are saying we reject more now. If you insist the new approach is better it would be good to find out why this is the case. I haven't looked into it but the only explanation that would make sense to me is that with the approach we are using in master now, we add these supernets from IRR which are usually shorter and add larger areas of coverage for a smaller number of lines. This leads to many potential entries from Routeviews being rejected which are longer than what we got from IRR but don't offer additional coverage. Since RV essentially is BGP traffic it makes sense that it is a lot more fractured than IRR which is used for filters etc. So after the code change here rejects these supernets from IRR this leaves the door open for all these longer routeviews entries to get into the result file, which results in more entries.

This is just an educated guess, I have not looked at the data yet to verify if this is what happens.

kartograf/merge.py
jurraca: just changing the text, not a new log message. This is one of the messages we decided to keep.

kartograf/trie.py
jurraca: thanks, I removed this function but you are correct :)

jurraca:
> When we have overlapping coverage in our resulting asmap we use the most precise (longest) prefix for the ASN. 

thanks, you're right, I was overthinking it. Longest prefix match wins. I removed the clause and function that attempted to check for a supernet of the checked network. Now it's just subset checking, but doesn't need the root_net index as the previous merge process did, so it's more general. 

Watching repro run CI...

jurraca:
a diff against a run with this version still yields about a 15k diff, with mostly unassigned as with the previous run. Even after encoding with `--fill` the diffs are still about 7k. Looking into it.

kartograf/merge.py
fjahr: This line just looks like it's only changing the text but below on L133 the old log message still exists: https://github.com/asmap/kartograf/pull/123/commits/10108906a04cdfc2e55ad10e1b591cb3655ba4b7#diff-db4db3f74df857bf1194562bdeaa881b0f91197d4185f45345ca7eee6e2cf50eR133 So there are now two logs where there used to be one.

jurraca:
After looking further, I found the bug, documented [here](https://github.com/asmap/kartograf/issues/144). 

To reproduce, using the latest run https://github.com/bitcoin-core/asmap-data/issues/51 , you can run a reproduction run with this branch then run a diff between the result files. It's about 16k line diff file, with 16076 out of 16855 (95%)  in the diff are "previously unassigned". It's a bit misleading because the "unassigned" being reported in the diff are just getting rolled up into the larger network, and are not individually assigned in the trie-based result. 

The lines with previously assigned AS help show the issue. For example, there's the line: `12.0.224.0/21 AS7018 # was AS2386` so this network was assigned  `AS2386` in the master branch code and `AS7018` in this branch's trie-based code. In both runs, the network assignment comes from routeviews `collectors/pfx2asn_clean.txt` with the `AS3286` assignment. RPKI has `12.0.0.0/22`, and that is the base file. In both runs, IRR gives us `12.0.0.0/8 AS7018`. Because of the bug, this IRR assignment does not get merged in. The merge process should see the `12.0.0.0/8 AS7018` assignment, and judge that the `12.0.224.0/21` network is already covered by the IRR entry. In the trie case, this is the case. But in the original merge, this is not the case, resulting in the `12.0.224.0/21 AS2386` assignment, since it does not overlap with the base file's `12.0.0.0/22` entry.

Either way the trie impl is better if it caught this. While it would be nice to be able to "reproduce" any old ASmap, the correctness seems more valuable. 

added a small merge test to make this behavior explicit and to sanity-check myself; we can remove later if necessary.

fjahr:
Rebase needed here now

jurraca:
interesting test failure, I can't reproduce it locally, must be a flaky test due to our setup with generating random subnets. reviewing.

kartograf/merge.py
jurraca: removed [78edff8](https://github.com/asmap/kartograf/pull/123/commits/78edff857418bca317dfffb0b0f2729743db4105)

fjahr:
> must be a flaky test due to our setup with generating random subnets

Yeah, my best guess would be that the rpki nets and the irr nets contained a duplicate by chance. Then the check assumes that net is not part of the result but it would be, just coming out of rpki, not irr.

kartograf/trie.py
```diff
@@ -71,6 +76,47 @@ def lookup(self, ip):
 
         return last_asn
 
+    def _lookup_network(self, network):
+        """Lookup an IP network and check for overlap with existing networks.
+
+        For RPKI-based merging, we consider a network 'included' if it overlaps
+        with any RPKI network (exact match, subset, or superset). We only want
```
fjahr: That's exactly what you fixed separately, right?

```suggestion
        with any RPKI network (exact match or subset). We only want
```

tests/test_merge.py
```diff
@@ -125,49 +141,3 @@ def test_merge_rpki_supersedes_irr_subnets(tmp_path):
     assert set(final_ips).isdisjoint(set(irr_ips)), (
         f"IRR subnets should not appear in merged output: {set(final_ips) & set(irr_ips)}"
     )
-
-def test_more_specific_base_does_not_suppress_extra_supernet(tmp_path):
```
fjahr, line 130: I don't think this test should be removed?

kartograf/trie.py
```diff
@@ -71,6 +76,47 @@ def lookup(self, ip):
...
+            max_bits = 128
+
+        addr_int = int(network.network_address)
+        prefix_len = network.prefixlen
+
+        node = root
+        last_asn = None
+
+        # Traverse the trie to find networks that contain this candidate network
+        for i in range(max_bits):
```
fjahr: This should probably use this to not spend extra time looping that will not give us a result

```suggestion
        for i in range(prefix_len)
```

kartograf/merge.py
```diff
@@ -1,16 +1,12 @@
...
-import os
 import shutil
-from types import SimpleNamespace
 import pandas as pd
 
 from kartograf.timed import timed
+from kartograf.trie import IPTrie
 from kartograf.util import get_root_network
 
-
```
fjahr, line 13: This blank line shouldn't be removed, it's typical style to have two blank lines above classes in python, even though our linter doesn't flag it.

kartograf/merge.py
```diff
@@ -21,64 +17,31 @@ class BaseNetworkIndex:
...
-    def contains_row(self, row):
-        root_net = row.PFXS_LEADING
-        version = ipaddress.ip_network(row.PFXS).version
-        if version == 4 and (root_net in self._v4_keys):
-            return self.check_inclusion(row, root_net, version)
-        if version == 6 and (root_net in self._v6_keys):
-            return self.check_inclusion(row, root_net, version)
+        try:
+            candidate = ipaddress.ip_network(row.PFXS, strict=False)
```
fjahr: Why should we change this to `strict=False`? I think we should keep that behavior.

kartograf/merge.py
```diff
@@ -21,64 +17,31 @@ class BaseNetworkIndex:
...
-        self._dict = {4: {}, 6: {}}
-        self._v4_keys = self._dict[4].keys()
-        self._v6_keys = self._dict[6].keys()
+        self._trie = IPTrie()
 
-    def update(self, pfx):
+    def update(self, pfx, asn):
         try:
-            ipn = ipaddress.ip_network(pfx)
+            ipn = ipaddress.ip_network(pfx, strict=False)
```
fjahr: Same here as the other comment, probably should keep it strict unless I misunderstand why this change is made.

kartograf/trie.py
```diff
@@ -44,9 +44,14 @@ def insert(self, network, asn):
         node.asn = asn
 
     def lookup(self, ip):
-        if not isinstance(ip, (ipaddress.IPv4Address, ipaddress.IPv6Address)):
-            raise TypeError("lookup expects an ip_address object")
-
+        if isinstance(ip, (ipaddress.IPv4Network, ipaddress.IPv6Network)):
+            return self._lookup_network(ip)
+        if isinstance(ip, (ipaddress.IPv4Address, ipaddress.IPv6Address)):
+            return self._lookup_address(ip)
```
fjahr: Not sure if this really needs to be under one lookup function. Can't we just have two separate functions and let the call side use the right one? It seems the network part could be more like `covers(network)` as better naming? And then the IP one could just stay using `lookup(ip)`.

kartograf/merge.py
```diff
@@ -1,16 +1,12 @@
...
 import shutil
-from types import SimpleNamespace
 import pandas as pd
 
 from kartograf.timed import timed
+from kartograf.trie import IPTrie
 from kartograf.util import get_root_network
 
-
 class BaseNetworkIndex:
```
fjahr, line 10: The docstring of this class is outdated, it still talks about the dict. But I think we don't even need it anymore, it just wraps the `IPTrie` class and nothing else so we can just use an instance of the `IPTrie` class for all this, moving the functions there or making them free?

tests/test_trie.py
```diff
@@ -267,3 +267,17 @@ def test_lookup_rejects_invalid_string():
 
     with pytest.raises(TypeError):
         trie.lookup("not.an.ip.address")
+
+
+def test_lookup_network():
```
fjahr: Could you add some IPv6 coverage here?

kartograf/merge.py
```diff
@@ -149,64 +112,26 @@ def extra_file_to_df(extra_file_path):
...
+            base_network_index.update(pfx, asn.strip())
 
     df_extra = extra_file_to_df(extra_file)
 
-    len_df_extra = len(df_extra)
-    chunk_size = pick_chunk_size(len_df_extra)
-    chunks = []
-    chunk_data = []
-    for i, row in df_extra.iterrows():
-        chunk_data.append((i, (row.INETS, row.PFXS, row.PFXS_LEADING)))
```
fjahr, line 193: It seems to me that `PFXS_LEADING` and `INETS` are not used for anything anymore after this change so I think you can remove them? Maybe all this will be obsolete if we get rid of pandas but not 100% sure about it and this might save us some memory and CPU cycles.

fjahr (commented):
Did another pass, could you squash the fixups please so the commit structure is a bit nicer?

I hope we can move forward with this after the run next week and then maybe look into removing the dependencies afterwards.

> Yeah, my best guess would be that the rpki nets and the irr nets contained a duplicate by chance. Then the check assumes that net is not part of the result but it would be, just coming out of rpki, not irr.

Pretty convinced that this is the case here. It's not a change that this pull introduces so I opened a fix in a separate pull so we can ignore this issue here from now: #155

kartograf/trie.py
jurraca: yea, I went with a different impl that has `lookup(ip)` , `covering_asn(network)`, and `insert(network)` all share an underlying `walk` function to walk the trie, see https://github.com/asmap/kartograf/pull/123/commits/de55b8be8bb630918cfc88655f9b74a0006d4ad9

kartograf/merge.py
jurraca: conceptually yes, but there's some behavior changes, so i'd prefer to do this as a followup.

jurraca:
Addressed feedback, rebased, and reorganized commits to be easier to read:
- trie implementation and tests https://github.com/asmap/kartograf/pull/123/commits/de55b8be8bb630918cfc88655f9b74a0006d4ad9
- use the trie in merge.py https://github.com/asmap/kartograf/pull/123/commits/8bc80766a2d5023127c01e37aee3f464eaaee701
- small line parsing fix https://github.com/asmap/kartograf/pull/123/commits/20da90f36681b4b400031971c31f125b4fa5dc61
- drop dead columns INETS/PFXS_LEADING https://github.com/asmap/kartograf/pull/123/commits/67332ec742c02d90d754eddacd660e05b11b6a36

Keeping removal of the `BaseNetworkIndex` class as a followup and further simplication. Thanks @fjahr !

kartograf/merge.py
```diff
@@ -121,92 +83,51 @@ def merge_pfx2as(context):
...
-
-
 def general_merge(
     base_file, extra_file, extra_filtered_file, out_file
 ):
     """
     Merge lists of IP networks into a base file.
     """
-    print("Merging extra prefixes that were not included in the base file.")
+    print("Creating network index from base file.")
```
fjahr, line 114: Having two logging here seems to be a regression from some of the reorgs, this log was removed earlier: https://github.com/asmap/kartograf/pull/123#discussion_r2886261020

fjahr (commented):
ACK 35d8963a33268111ce94cdd3ea7a4fb1450be069

Thanks for addressing my feedback. Aside from re-reviewing, I compared hash results of 4 separate runs between this pull and master, the hashes matched each time. The performance was consistently better and saved a little over minute for each run (almost 20%), memory usage was usually better too but I got one odd run where the peak was almost the same.

Great work!

kartograf/merge.py
fjahr: Maybe fix this when this code is touched the next time or in a small follow-up.

PR #125, opened 2026-03-05
followups from #116: handle divide by zero, remove timed

jurraca:
Return early if the IP list is empty.
Also, moves the trie creation after checking the IP list, since it is time-consuming.
Adds an empty IP list test.

I kept the "Running coverage check" print statement because, while the command is faster, it can still take around 10 seconds on my machine, which is a little long for having no feedback for the user. I removed the newline though. The statement is only printed after the IP list is checked and before the trie starts being built.

### DIFF
--- a/kartograf/coverage.py
+++ b/kartograf/coverage.py
@@ -1,13 +1,7 @@
 import ipaddress
 from kartograf.trie import IPTrie
-from kartograf.timed import timed
 
-@timed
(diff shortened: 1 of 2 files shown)


fjahr (commented):
utACK 2fd9b758f0aa8b2d760c38fb2de73b563f9de197

Thanks for addressing it quickly!

Issue #96, opened 2025-10-05
rpki-client >= 9.6 `CCR hashes`


job:
Ola!

Rpki-client 9.6 has a new feature that seems useful for the kartograf project: `CCR manifest state hash` and `CCR ROA payloads hash` (displayed in the JSON output and as log message on the console at the end of the program's execution).

CCR stands for Canonical Cache Representation: https://datatracker.ietf.org/doc/html/draft-spaghetti-sidrops-rpki-ccr

The core idea is that by using a deterministic canonical way to represent the state of the validated cache, it is easier to compare the state of different validator instances with each other. (different validator runs across points in time or validators on different computers)

The two hashes that seem relevant for the project are the `CCR manifest hash` and `ROA payload hash`. The first one is the head of a merkle tree that represents the entire "physical" state of all the files in the validated cache directory. The second (ROA) hash is a hash over the deduplicated Prefix<>ASN tuples that are the payload of the ROAs.

If a RPKI CA resigns a CRL but doesn't change any ROAs, then the Manifest hash will change but not the ROA hash. If a RPKI CA adds or removes a prefix entry in a ROA then both hashes will change. These hashes should make it easier to recognize whether the "form" or the "substance" of the RPKI changed between 2 runs.

fjahr:
Awesome, that looks very intersting for us! We have been testing rpki-client 9.6 with kartograf and @jurraca is experimenting with the `-p` option which looks to work very well so far! But we hadn't noticed the significance this feature, we'll definitely look into it!

@jurraca are you interested in this or should I pick this up while you work on `-p` and the other stuff you have in the pipeline?

PR #127, opened 2026-04-05
Print CCR hashes in run logs

fjahr:
This is an experiment with the suggestion from #96. This extracts the CCR hashes from `rpki-client` output. These hashes are potentially nicer to compare rpki cache contents because they deduplicate automatically for example.

First downside is that we need to raise the minimum `rpki-client` version to 9.6 because older versions don't have these CCR hashes. Second downside is that these are not being calculated in file mode in a way that is comparable afaict so we seem to need to do duplicate work in the case of reproduction runs. We can probably do this smarter and just get rid of file mode usage or maybe there is a way to get it to work in file mode, but I haven't spent enough time on this yet.

I have only done shallow testing so far. At least with the additional `-P` params the hashes have matched between the reproduction runs and the original downloads. Maybe that's making it worth to check the results on the latest run which has failed but I am not sure how actionable the results would be.

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,7 +56,7 @@ podman rm kartograf-run
 
 #### rpki-client
 
-Kartograf requires `rpki-client` version 8.4 or higher to be installed locally. Many package managers have `rpki-client`, however please note that typically only the latest version is available. Kartograf is currently tested to work with `rpki-client` 8.4 - 9.7. If a new release is available that breaks compatibility, you may need to build a compatible `rpki-client` version yourself if your preferred package manager does not have a compatible version available. In that case, see the [instructions in the rpki-client-portable project](https://github.com/rpki-client/rpki-client-portable/blob/master/INSTALL).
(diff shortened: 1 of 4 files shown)


jurraca:
concept ACK, it makes sense to leverage this CCR representation.

However I can't seem to get any output. Every run has empty `result.stdout`, so nothing is printed in output or log file. Am I holding it wrong?

regarding file mode, if you mean using the -f flag, then #97 is relevant since it removes that flag (we decided to park the change for now).

fjahr:
> However I can't seem to get any output. Every run has empty result.stdout, so nothing is printed in output or log file. Am I holding it wrong?

In `nix` right, I only tested without it, will investigate.

> regarding file mode, if you mean using the -f flag, then https://github.com/asmap/kartograf/pull/97 is relevant since it removes that flag (we decided to park the change for now).

yeah, I looked into it again but I still think that making the reproduction run a bit slower is still better than changing our decision making. I hope they can fix this omitting of the values we are using in the context of `-p` on the `rpki-client` side...

fjahr:
> However I can't seem to get any output. Every run has empty result.stdout, so nothing is printed in output or log file. Am I holding it wrong?

I think I have fixed it, can you check if it works for you now? I will still need to spend a bit more time on it and make sure this is the way to do it.

jurraca:
ACK [43f86d7](https://github.com/asmap/kartograf/pull/127/commits/43f86d7c6546384de639cd870acc8855116e0feb) 
yes, it works now. Thanks for adding the explanation.

fjahr:
I had to fix a typo in the readme that I introduced, so I think this is mergable. I didn't find a better approach for printing those hashes and it would be good to see them in action soon.

PR #129, opened 2026-04-07
Use epoch_datetime for RouteViews month selection

### LINKED ISSUE #128: RouteViews fetch uses wall-clock time, causing inconsistent data in coordinated runs
During coordinated launches, RouteViews URLs are resolved after RPKI/IRR fetching. Since participants finish at different times, they can see different CAIDA directory states as observed in bitcoin-core/asmap-data#46 where the April directory appeared mid-run.

Fix: resolve RouteViews URLs at run start, use `epoch_datetime` for month selection, and fall back on empty directories.

pradhyum6144:
Resolve RouteViews URLs at run start (before RPKI/IRR) so all coordinated-launch participants see the same CAIDA directory state. Also uses epoch_datetime for month selection and falls back on empty directories.

Fixes #128, observed in bitcoin-core/asmap-data#46.

### DIFF
--- a/kartograf/collectors/routeviews.py
+++ b/kartograf/collectors/routeviews.py
@@ -85,15 +85,24 @@ def extract(file, context):
             write.write(formatted + '\n')
 
 
+def resolve_routeviews_urls(context):
+    """Resolve RouteViews download URLs for this epoch."""
+    epoch_dt = context.epoch_datetime
+    context.routeviews_v4_url = latest_link(PFX2AS_V4, epoch_dt)
+    context.routeviews_v6_url = latest_link(PFX2AS_V6, epoch_dt)
+    print(f"Resolved RouteViews URLs:\n  v4: {context.routeviews_v4_url}"
+          f"\n  v6: {context.routeviews_v6_url}")
+
+
 @timed
 def fetch_routeviews_pfx2as(context):
     path = Path(context.data_dir_collectors)
     v4_file_gz = path / "routeviews_pfx2asn_ip4.txt.gz"
     v6_file_gz = path / "routeviews_pfx2asn_ip6.txt.gz"
 
-    download(latest_link(PFX2AS_V4, context.epoch_datetime), v4_file_gz)
+    download(context.routeviews_v4_url, v4_file_gz)
     print(f"Downloaded {v4_file_gz.name}, file hash: {calculate_sha256(v4_file_gz)}")
-    download(latest_link(PFX2AS_V6, context.epoch_datetime), v6_file_gz)
+    download(context.routeviews_v6_url, v6_file_gz)
     print(f"Downloaded {v6_file_gz.name}, file hash: {calculate_sha256(v6_file_gz)}")
 
 
--- a/kartograf/collectors/routeviews.py
+++ b/kartograf/collectors/routeviews.py
@@ -1,4 +1,4 @@
-from datetime import datetime, timedelta
+from datetime import timedelta
 from pathlib import Path
 import gzip
 import shutil
@@ -16,9 +16,8 @@ PFX2AS_V4 = "https://publicdata.caida.org/datasets/routing/routeviews-prefix2as/
 PFX2AS_V6 = "https://publicdata.caida.org/datasets/routing/routeviews6-prefix2as/"
 
 
-def latest_link(base):
-    now = datetime.now()
-    ym = year_and_month(now)
+def latest_link(base, epoch_datetime):
+    ym = year_and_month(epoch_datetime)
     url = base + ym
 
     try:
@@ -28,7 +27,7 @@ def latest_link(base):
         print(f"The page at {url} couldn't be fetched. "
               f"Trying the previous month.")
 
-        last_month = now - timedelta(days=now.day)
+        last_month = epoch_datetime - timedelta(days=epoch_datetime.day)
         ym = year_and_month(last_month)
         url = base + ym
 
@@ -92,9 +91,9 @@ def fetch_routeviews_pfx2as(context):
     v4_file_gz = path / "routeviews_pfx2asn_ip4.txt.gz"
     v6_file_gz = path / "routeviews_pfx2asn_ip6.txt.gz"
 
-    download(latest_link(PFX2AS_V4), v4_file_gz)
+    download(latest_link(PFX2AS_V4, context.epoch_datetime), v4_file_gz)
     print(f"Downloaded {v4_file_gz.name}, file hash: {calculate_sha256(v4_file_gz)}")
-    download(latest_link(PFX2AS_V6), v6_file_gz)
+    download(latest_link(PFX2AS_V6, context.epoch_datetime), v6_file_gz)
     print(f"Downloaded {v6_file_gz.name}, file hash: {calculate_sha256(v6_file_gz)}")
 
 
--- a/kartograf/collectors/routeviews.py
+++ b/kartograf/collectors/routeviews.py
@@ -8,7 +8,7 @@ from bs4 import BeautifulSoup
 import requests
 
 from kartograf.timed import timed
-from kartograf.util import calculate_sha256
+from kartograf.util import calculate_sha256, download_with_retries
 
 # Routeviews Prefix to AS mappings Dataset for IPv4 and IPv6
 # https://www.caida.org/catalog/datasets/routeviews-prefix2as/
@@ -20,35 +20,36 @@ def latest_link(base, epoch_datetime):
     ym = year_and_month(epoch_datetime)
     url = base + ym
 
-    try:
-        response = requests.get(url, timeout=600)
-        response.raise_for_status()
-    except requests.exceptions.HTTPError:
-        print(f"The page at {url} couldn't be fetched. "
-              f"Trying the previous month.")
+    response = download_with_retries(url)
+    if response:
+        latest = _pick_latest_pfx2as(response.text)
+        if latest:
+            return url + latest
 
-        last_month = epoch_datetime - timedelta(days=epoch_datetime.day)
-        ym = year_and_month(last_month)
-        url = base + ym
+    print(f"No pfx2as.gz files at {url}. Trying the previous month.")
 
-        try:
-            response = requests.get(url, timeout=600)
-            response.raise_for_status()
-        except requests.exceptions.HTTPError:
-            print(f"The page at {url} couldn't be fetched. "
-                  f"Download of Routeviews pfx2as data failed.")
-            sys.exit()
+    last_month = epoch_datetime - timedelta(days=epoch_datetime.day)
+    fallback_url = base + year_and_month(last_month)
 
-    soup = BeautifulSoup(response.text, 'html.parser')
+    response = download_with_retries(fallback_url)
+    if response:
+        latest = _pick_latest_pfx2as(response.text)
+        if latest:
+            return fallback_url + latest
 
+    print(f"The page at {fallback_url} also has no pfx2as.gz files. "
+          f"Download of Routeviews pfx2as data failed.")
+    sys.exit()
+
+
+def _pick_latest_pfx2as(html):
+    soup = BeautifulSoup(html, 'html.parser')
     links = [a["href"] for a in soup.find_all("a", href=True)]
     latest = ""
-
     for link in links:
         if link.endswith(".pfx2as.gz"):
             latest = link
-
-    return url + latest
+    return latest
 
 
 def year_and_month(now):
--- a/kartograf/kartograf.py
+++ b/kartograf/kartograf.py
@@ -7,7 +7,11 @@ from kartograf.cleanup import cleanup_out_files
 
 from kartograf.context import Context
 from kartograf.coverage import coverage
-from kartograf.collectors.routeviews import extract_routeviews_pfx2as, fetch_routeviews_pfx2as
+from kartograf.collectors.routeviews import (
+    extract_routeviews_pfx2as,
+    fetch_routeviews_pfx2as,
+    resolve_routeviews_urls,
+)
 from kartograf.collectors.parse import parse_routeviews_pfx2as
 from kartograf.irr.fetch import extract_irr, fetch_irr
 from kartograf.irr.parse import parse_irr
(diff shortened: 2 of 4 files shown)


pradhyum6144:
@fjahr @jurraca This addresses the RouteViews race condition observed in the April 2 coordinated run (discussed in #46). Would appreciate a review when you get a chance.

jorisstrakeljahn:
The empty-directory check (`if not latest: raise ValueError`) is a nice addition for the first attempt, but the same validation is missing in the fallback path:

```python
# First attempt — validated:
latest = _pick_latest_pfx2as(response.text)
if not latest:
    raise ValueError("no pfx2as.gz files in directory")

# Fallback — not validated:
latest = _pick_latest_pfx2as(response.text)
return url + latest  # if latest is "", this returns a bare directory URL
```

If the previous month's directory also has no `.pfx2as.gz` files, `download()` would save the HTML directory listing as a gzip file. Adding the same check here would make the fix complete. This was also an issue in the original code, so it would be good to cover both paths.

Also, one thing about the core problem from the April 2 launch (bitcoin-core/asmap-data#46): in a coordinated run, `epoch_datetime` and `datetime.now()` return the same month since everyone runs on the same day. The divergence wasn't caused by targeting different months — all participants tried `/2026/04/`. It was caused by the directory not existing yet at 16:00 UTC and only appearing around 16:20. The `epoch_datetime` change helps with non-coordinated/reproduction runs, but a retry before falling back might be needed to address that specific race condition.

pradhyum6144:
> The empty-directory check (`if not latest: raise ValueError`) is a nice addition for the first attempt, but the same validation is missing in the fallback path:
> 
> ```python
> # First attempt — validated:
> latest = _pick_latest_pfx2as(response.text)
> if not latest:
>     raise ValueError("no pfx2as.gz files in directory")
> 
> # Fallback — not validated:
> latest = _pick_latest_pfx2as(response.text)
> return url + latest  # if latest is "", this returns a bare directory URL
> ```
> 
> If the previous month's directory also has no `.pfx2as.gz` files, `download()` would save the HTML directory listing as a gzip file. Adding the same check here would make the fix complete. This was also an issue in the original code, so it would be good to cover both paths.
> 
> Also, one thing about the core problem from the April 2 launch ([bitcoin-core/asmap-data#46](https://github.com/bitcoin-core/asmap-data/issues/46)): in a coordinated run, `epoch_datetime` and `datetime.now()` return the same month since everyone runs on the same day. The divergence wasn't caused by targeting different months — all participants tried `/2026/04/`. It was caused by the directory not existing yet at 16:00 UTC and only appearing around 16:20. The `epoch_datetime` change helps with non-coordinated/reproduction runs, but a retry before falling back might be needed to address that specific race condition.

Thanks both addressed in the latest push:

Fallback is now validated  extracted into _try_fetch_latest() which returns None if no .pfx2as.gz files exist. Both paths use it and an empty fallback exits with a clear error.

Added retry (3 attempts, 120s apart) before falling back to previous month. Would have caught the April 2 race where the directory appeared ~20min after launch.

jorisstrakeljahn:
The retry is a good addition, one thing worth considering though: with the retry happening before RPKI/IRR, every coordinated launch near a month boundary would stall for up to 4 minutes (3 attempts, 2 sleeps of 120s) before RPKI even starts. In the April 2 case, the last retry would have been around 16:04 UTC, the directory appeared between 16:13 and 16:21, so the retries would have still missed it and fallen back to March anyway.

The real fix here is the early resolution, all participants resolve at the same time and get the same result regardless of download speed. The retry is a bonus for cases where the timing gap is smaller. But the blocking `time.sleep(120)` with no progress feedback might look like the program is hanging, especially since nothing prints for 2 minutes.

A few options to consider:
- Shorter delay with more attempts (e.g. 30s x 8) — same total wait, but more responsive
- Only retry in coordinated launch mode (-w) since non-coordinated runs don't need consistency
- Print a countdown or periodic status during the wait

jurraca:
> The real fix here is the early resolution

yes agreed, the retry is unnecessary or should be very limited. It should block execution as little as possible.

> Only retry in coordinated launch mode (-w) since non-coordinated runs don't need consistency

the `-w` flag is really just a set timer to launch feature, so while it's convenient for collaborative launches, it shouldn't imply anything about file consistency.

pradhyum6144:
Agreed, removed the retry loop. The early resolution ensures all participants see the same CAIDA state the retry added blocking time for minimal benefit. Now it's a single attempt with validated fallback on both paths.

jurraca:
sorry @pradhyum6144 I should have been more clear. We should still be retrying the request if it fails. The problem with [62d16d2](https://github.com/asmap/kartograf/pull/129/commits/62d16d2942d2e499cb7a4f9d388b9b3e90f8847f) is that you're retrying for a URL which may not have any data (i.e. a folder which doesn't have any routeviews file to download). So if it's absent you would spend 120 seconds retrying something that isn't there. 

You should draw a distinction between the request succeeding, but no data is there for us (in which case we look for the previous month), and the request failing where we need to retry. 

@jorisstrakeljahn lmk if this matches your understanding.

jorisstrakeljahn:
Yes, that matches my understanding. The two cases are:

1. Request succeeds (200) but no `.pfx2as.gz` files — the directory exists but CAIDA hasn't uploaded the data yet. No point retrying, just fall back to the previous month immediately.
2. Request fails (network error, timeout, HTTP 5xx) — transient issue, should be retried.

Right now `_try_fetch_latest` returns `None` for both cases, so the caller can't tell them apart. It could return the URL on success, `None` for "empty directory", and let the exception propagate for actual request failures, that way the retry logic only kicks in when the request itself failed, not when the data just isn't there yet.

pradhyum6144:
Updated. _try_fetch_latest now distinguishes between the two cases:

- Request succeeds but no data (200 OK, no .pfx2as.gz files) → returns None, falls back to previous month immediately
- Request fails (network error, timeout, 5xx) → retries up to 3 times with 30s delay before raising

The retry logic is isolated in a new _fetch_with_retry wrapper so _try_fetch_latest stays clean  it just returns the URL, None, or lets the exception propagate.

image


image


image

kartograf/collectors/routeviews.py
```diff
@@ -15,41 +16,69 @@
...
-        print(f"The page at {url} couldn't be fetched. "
-              f"Trying the previous month.")
+def _try_fetch_latest(base_url):
+    """Fetch directory listing and return the latest pfx2as file.
+
+    Returns the full URL on success, None if the page exists but contains
+    no pfx2as.gz files.  Raises on request failures so the caller can retry.
+    """
+    response = requests.get(base_url, timeout=600)
+    response.raise_for_status()
```
jorisstrakeljahn: raise_for_status() raises HTTPError for all non-2xx responses, including 404. Since HTTPError is a subclass of RequestException, _fetch_with_retry catches and retries 404s. After all retries exhaust, the exception propagates out of latest_link, the fallback to the previous month is never reached.

In the April 2 scenario, CAIDA returned 404 for /2026/04/. With this code, that would retry 3 times and crash instead of falling back to March.

A 404 means "no data for this month" (same as an empty directory) and should return None:

```suggestion
    if response.status_code == 404:
        return None
    response.raise_for_status()
```

If this is addressed, the rest looks good to me @jurraca

kartograf/collectors/routeviews.py
```diff
@@ -15,41 +16,69 @@
 PFX2AS_V4 = "https://publicdata.caida.org/datasets/routing/routeviews-prefix2as/"
 PFX2AS_V6 = "https://publicdata.caida.org/datasets/routing/routeviews6-prefix2as/"
 
+RETRY_ATTEMPTS = 3
+RETRY_DELAY = 30
```
jurraca: 10 seconds is probably fine

pradhyum6144:
@jorisstrakeljahn @jurraca Addressed both 404 now returns None (falls back without retrying) and retry delay reduced to 10s.

pradhyum6144:
@jorisstrakeljahn @jurraca,  @achow101 suggested on #128 that we could also filter files by their filename timestamp so _pick_latest_pfx2as only picks files that existed by the epoch time. Should I add that as a follow-up PR after this one is merged?

jorisstrakeljahn (approved):

fjahr:
Hi @pradhyum6144 , thanks for working on this. Can you please squash the fixup commits and separate the changes by topic? Currently this is a bit hard to review. Maybe keep the first commit for the epoch, then one for resolving the URLs early and then one for the retry stuff.

kartograf/collectors/routeviews.py
fjahr: We already have retry logic in `irr/fetch.py`, I don't think we necessarily need to have different logic between IRR and routeviews, so it would be good to deduplicate this and have a `download_with_retries` util function or something.

kartograf/collectors/routeviews.py
```diff
@@ -40,16 +42,19 @@ def latest_link(base):
                   f"Download of Routeviews pfx2as data failed.")
             sys.exit()
 
-    soup = BeautifulSoup(response.text, 'html.parser')
+        latest = _pick_latest_pfx2as(response.text)
+
+    return url + latest
+
 
+def _pick_latest_pfx2as(html):
```
fjahr, line 45: It doesn't seem like it's necessary to introduce this helper function?

kartograf/kartograf.py
```diff
@@ -7,7 +7,7 @@
 
 from kartograf.context import Context
 from kartograf.coverage import coverage
-from kartograf.collectors.routeviews import extract_routeviews_pfx2as, fetch_routeviews_pfx2as
+from kartograf.collectors.routeviews import extract_routeviews_pfx2as, fetch_routeviews_pfx2as, resolve_routeviews_urls
```
fjahr: nit: Could start to do the import multi-line here

kartograf/collectors/routeviews.py
```diff
@@ -91,16 +91,25 @@ def extract(file, context):
             write.write(formatted + '\n')
 
 
+def resolve_routeviews_urls(context):
+    """Resolve RouteViews download URLs early so all participants in a
```
fjahr: This function is agnostic to whether it is called early or not. So only documenting that it is called early at the call-site is better.

kartograf/collectors/routeviews.py
```diff
@@ -91,16 +91,25 @@ def extract(file, context):
...
+    coordinated launch see the same CAIDA directory state."""
+    epoch_dt = context.epoch_datetime
+    context.routeviews_v4_url = latest_link(PFX2AS_V4, epoch_dt)
+    context.routeviews_v6_url = latest_link(PFX2AS_V6, epoch_dt)
+    print(f"Resolved RouteViews URLs:\n  v4: {context.routeviews_v4_url}"
```
fjahr, line 94: This prints kind of randomly in the logs, maybe introduce a section header before this calling this phase "Pre-fetch" or something (I can't come up with a good name right now :)

kartograf/collectors/routeviews.py
fjahr: We definitely want to move on immediately if the page doesn't exist. If we needlessly loop through retries even for a few seconds here we increase the chance of diverging in the RPKI results and then we may introduce more issues with this change than we are fixing 😓

fjahr:
@pradhyum6144 Can you please notify us with a comment when you have addressed and resolve review comments? Otherwise I don't know when this is ready for me to take another look or if you are still working on it. I let the CI run and it seems the linter has some complaints.

kartograf/collectors/routeviews.py
fjahr: I would still prefer it if this style wasn't changed here. I don't see a need for it and I think it results in more code and is a bit harder to parse as a result.

kartograf/collectors/routeviews.py
```diff
@@ -20,35 +20,36 @@ def latest_link(base, epoch_datetime):
...
-            response.raise_for_status()
-        except requests.exceptions.HTTPError:
-            print(f"The page at {url} couldn't be fetched. "
-                  f"Download of Routeviews pfx2as data failed.")
-            sys.exit()
+    last_month = epoch_datetime - timedelta(days=epoch_datetime.day)
+    fallback_url = base + year_and_month(last_month)
 
-    soup = BeautifulSoup(response.text, 'html.parser')
+    response = download_with_retries(fallback_url)
```
fjahr, line 34: `download_with_retries` with retries in now used for the picking of the URL but not for the actual download. Would be good to use them for both.

kartograf/collectors/routeviews.py
```diff
@@ -20,35 +20,36 @@ def latest_link(base, epoch_datetime):
     ym = year_and_month(epoch_datetime)
     url = base + ym
 
-    try:
-        response = requests.get(url, timeout=600)
-        response.raise_for_status()
-    except requests.exceptions.HTTPError:
-        print(f"The page at {url} couldn't be fetched. "
-              f"Trying the previous month.")
+    response = download_with_retries(url)
```
fjahr, line 23: We should lower the timeout here if we do this early before RPKI. Probably more like 60s, if we are hanging here for a while we won't get a match on RPKI anyway so we might as well fail right away if can't fetch such a simple page.

fjahr (commented):
The first two commits look good to me but I am not sure about the last one with the retries. It seems the CI failure that is not addressed yet is also caused by the third commit. I would really like to merge the first two commits short terms so we can have them for our next run this week.

@pradhyum6144 could you please drop the third commit here and move it into a separate PR, addressing the remaining comments? If you could do that I think we can move forward with these two commits quickly.

kartograf/collectors/routeviews.py
jurraca: i'd rather omit this log message, not particularly interesting to the user imo

jurraca (commented):
first two commits are good, but second fjahr's comments about the third.

kartograf/collectors/routeviews.py
fjahr: I tend to agree, we are still printing the URLs when we actually download.

PR #130, opened 2026-04-07
Add RPKI diff tool for comparing collaborative launch data

pradhyum6144:
Following the discussion in bitcoin-core/asmap-data#46 where @jurraca mentioned needing "a more complete upload and diff engine to make sense of these runs", this adds a tool to compare RPKI data between participants at three levels: cache directory hashes, ROA-level diffs, and final map impact.

Tested on @achow101's uploaded SF vs NY data from the April 2 run. The cache-level diffs look large (2,220 extra ROAs, 5/15 repos differ), but only 4 out of 709,858 prefixes actually diverge in the final RPKI map. The larger hash divergence across participants likely comes from compounding with IRR and RouteViews differences rather than RPKI alone.


pradhyum6144:
@jurraca @fjahr This follows up on the discussion in bitcoin-core/asmap-data#46 about needing better tooling to compare RPKI data across participants.

I tested it on @achow101's uploaded SF vs NY data from the April 2 collaborative launch. Here's what I found:
image


image






image


Cache-level diffs look large (5/15 repos differ, 2,220 extra ROAs) but only 4 out of 709,858 prefixes actually diverge in the final RPKI map. The bigger hash divergence in collaborative launches likely comes from IRR and RouteViews compounding on top of small RPKI differences.

fjahr:
Hi @pradhyum6144 , thanks for working on this. Some high level observations: I like that this can filter the actual impact of the difference to the output, that's worth focussing on and showing more prominently (the 4 prefixes you mention). I think for the cache/roa diff it would make sense to use CCR hashes generated from `rpki-client`, those should be more informative and we shouldn't try to re-invent the wheel there, see #127. The dependency issues isn't so much a concern if this is a standalone script like you have done it now. You will have to check how performance is though. Also, I think this script should only focus on comparing the cache directory and included roas. We already have a diff command for maps in asmap-tool and I don't think I understand what the `parse-hashes` command is doing.

pradhyum6144:
I'll remove compare-maps since asmap-tool already covers that and drop parse-hashes  it was a helper to compare per-repo hash outputs from participants logs, but that's not needed if we use CCR hashes from rpki-client. I'll look into using CCR hashes from #127 for the cache comparison instead of custom directory hashing and focus the tool on cache directory + ROA comparison with emphasis on showing the actual prefix-level impact. Will check performance with CCR hashes and update the PR.

PR #131, opened 2026-04-08
Filter RouteViews pfx2as files by epoch timestamp

pradhyum6144:
Follow-up to #129. As suggested by @achow101 in #128, `_pick_latest_pfx2as` now parses the timestamp from pfx2as filenames (e.g. `routeviews-rv2-20260402-1600.pfx2as.gz`) and skips any file with a timestamp after the epoch. This ensures coordinated-launch participants select the same file even when new files appear on CAIDA mid-run.

### DIFF
--- a/kartograf/collectors/routeviews.py
+++ b/kartograf/collectors/routeviews.py
@@ -1,9 +1,9 @@
-from datetime import datetime, timedelta, timezone
+from datetime import timedelta
 from pathlib import Path
 import gzip
-import re
 import shutil
 import sys
+import tempfile
 import time
 
 from bs4 import BeautifulSoup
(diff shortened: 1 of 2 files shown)


pradhyum6144:
image





image





image


Tested the timestamp filtering with simulated CAIDA directory listings. Files after the epoch are skipped, so coordinated-launch participants always pick the same file regardless of when new data appears mid-run.

pradhyum6144:
@achow101 @jurraca  @jorisstrakeljahn @fjahr Implemented the timestamp filtering suggested in #128. _pick_latest_pfx2as now parses timestamps from filenames and skips any file after the epoch. Tested with simulated CAIDA listings files after the epoch are correctly filtered out.

fjahr:
Hi, I am not sure if the filename timestamp would make a difference since the file timestamp is always the one from the day before the upload and it varies kind of randomly independent of when the file is actually uploaded. The upload time would be much more interesting to check instead as belt and suspenders but you would need to normalize the timezone.

pradhyum6144:
I was assuming the filename timestamp reflected when the file appeared on CAIDA, but if it's the collection time from the day before then filtering on it won't help. I'll look into parsing the upload timestamps from the directory listing instead. Those are shown in the Apache index page next to each file, so _pick_latest_pfx2as could extract them from the HTML along with the filenames. Will need to check what timezone CAIDA uses and normalize accordingly.

pradhyum6144:
@fjahr i have Updated now parsing the upload timestamps from the Apache directory listing instead of the filename. CAIDA timestamps are in US/Pacific, so I convert to UTC using the PDT offset (UTC-7) as a conservative bound. Tested against the April 2 scenUpdated now parsing the upload timestamps from the Apache directory listing instead of the filename. CAIDA timestamps are in US/Pacific, so I convert to UTC using the PDT offset (UTC-7) as a conservative bound. Tested against the April 2 scenario: at epoch 16:00 UTC the file uploaded at 16:13 UTC is correctly skipped.ario: at epoch 16:00 UTC the file uploaded at 16:13 UTC is correctly skipped.

achow101:
If we're ok with downloading things twice, the timestamp itself is encoded in gzip header the gzip library can retrieve it. It might be better to do that than to parse the HTML to determine whether the downloaded file is usable.

fjahr:
> If we're ok with downloading things twice, the timestamp itself is encoded in gzip header the gzip library can retrieve it. It might be better to do that than to parse the HTML to determine whether the downloaded file is usable.

That's very interesting. It seems to be worth checking out and exploring that approach.

```
$ gzip -lNv routeviews-rv2-20260331-1200.pfx2as.gz
method  crc     date  time    compressed uncompressed  ratio uncompressed_name
defla 603a46ba Apr  1 18:13      3749393     23544660  84.0% routeviews-rv2-20260331-1200.pfx2as
```

fjahr:
@pradhyum6144 Same comment here as on the other PR, please comment when you have addressed review comments since otherwise I don't know if this is ready for another round of review.

fjahr:
@pradhyum6144 I have merged your first two commits separately in #140 , thank you for implementing these. Can you please still rebase here and address the feedback on the retries. I would also suggest that you address @jurraca 's comment on the logging in a separate commit. Thank you!

PR #132, opened 2026-04-09
test: fix test data generation bugs and improve merge test readability

jorisstrakeljahn:
Reviewed `test_merge.py` and found two bugs in `generate_data.py` that made tests less effective than intended.

Bug 1: `generate_subnets_from_base` used `str(subnet.network_address)` which returns bare addresses like `'10.0.0.0'` instead of CIDR notation `'10.0.0.0/17'`. The overlap assertion in `test_merge_joint` compared these against merge output entries like `'10.0.0.0/16'`. Since bare addresses never match CIDR strings, `isdisjoint` was always true regardless of whether the merge actually worked.

Bug 2: `generate_ip` used `'::::::::'` as the IPv6 zero address, which is invalid notation (`::` is correct — only one `::` is permitted per address). Every call with `ip_type="v6"` raised `AddressValueError`, so IPv6 test data generation has never worked. Nothing in the test suite exercised that path before, so the mistake stayed hidden until IPv6 coverage was added here.

Other changes:
- Rename `__tmp_paths`/`__read_test_vectors` to single underscore (name mangling is only meaningful inside classes)
- Rename `test_merge` → `test_merge_identical_files`, `test_merge_joint` → `test_merge_rpki_supersedes_irr_subnets`
- Parametrize `test_merge_identical_files` for v4/v6 instead of duplicating the test
- Add assertion message to the overlap test for better failure diagnostics

### DIFF
--- a/tests/test_merge.py
+++ b/tests/test_merge.py
@@ -15,10 +15,10 @@ from .util.generate_data import (
 )
 
 
-def __tmp_paths(tmp_path):
+def _tmp_paths(tmp_path):
     return [tmp_path / p for p in ["rpki_final.txt", "irr_final.txt", "out.txt"]]
 
-def __read_test_vectors(filepath):
+def _read_test_vectors(filepath):
     '''
     Fixtures for IP networks are under tests/data.
     Read them and return the list of valid networks and the list of individual subnets in the file.
@@ -41,9 +41,9 @@ def test_merge_from_fixtures(tmp_path):
     i.e. subnets are merged into the root network appropriately.
     '''
     data_dir = Path(__file__).parent / "data"
-    base_nets, base_results = __read_test_vectors(data_dir / "base_file.csv")
+    base_nets, base_results = _read_test_vectors(data_dir / "base_file.csv")
     base_path = tmp_path / "base.txt"
-    extra_nets, extra_results = __read_test_vectors(data_dir / "extra_file.csv")
+    extra_nets, extra_results = _read_test_vectors(data_dir / "extra_file.csv")
     extra_path = tmp_path / "extra.txt"
     # write the networks to disk, generating ASNs for each network
     generate_ip_file(base_path, build_file_lines(base_nets, generate_asns(len(base_nets))))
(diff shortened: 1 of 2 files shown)


jurraca:
great finds! No comments, tested and approved. Thanks!

jurraca (approved):

tests/test_merge.py
```diff
@@ -59,11 +61,12 @@ def test_merge_from_fixtures(tmp_path):
...
     assert merged_networks == expected_networks
 
-def test_merge_identical_files(tmp_path):
+@pytest.mark.parametrize("ip_type", ["v4", "v6"])
+def test_merge_identical_files(tmp_path, ip_type):
     '''
     Assert that merging two identical files is a no-op.
     '''
-    rpki_data = generate_file_items(100)
+    rpki_data = generate_file_items(100, ip_type=ip_type)
```
fjahr, line 68: Hm, keeping the default subnet range here makes this a bit unrealistic. The ipv6 range should be 16 to 48 if that is the ip_type used here, I think.

tests/test_merge.py
```diff
@@ -3,6 +3,8 @@
 '''
 from pathlib import Path
```
fjahr: nit: Don't need the newlines, we mix the imports usually, not sure why it's so spaced out in this file...

tests/test_merge.py
fjahr: Seems like we never set these params explicitly so far, so you could internalize the selection based on the ip type, I think. But no strong preference, you can also keep it flexible and set them from here.

fjahr (commented):
Looks good, but I would like the subnet range to change so it matches the ipv6 case.

tests/test_merge.py
jorisstrakeljahn: Done. Kept the params optional via None-sentinel so callers can still override.

tests/test_merge.py
jorisstrakeljahn: Removed

jorisstrakeljahn:
@fjahr, addressed review comments. Subnet-range defaults now picked from `ip_type` (8–24 for v4, 16–48 for v6) inside `generate_ip_networks`.

fjahr:
ACK 3e2568108770da80470d0183d4d3855454b8b2ab

Sorry for taking so long to get back around to this.

PR #137, opened 2026-04-18
merge/irr: harden parsing and deterministic output

SuvidhJ:
Small hardening follow-up from regression checks.

Main changes:
1) fix IRR entry splitting for blank-line handling and final-entry flush
2) make merge output writing deterministic and avoid blank-line artifacts
3) cap merge worker count for safer behavior on constrained systems
4) add deterministic merge regression coverage

Validation:
-> pytest tests/test_merge.py tests/test_irr_parse.py -q
-> pytest -q

Issue linkage:
- addresses #138

### DIFF
--- a/kartograf/irr/parse.py
+++ b/kartograf/irr/parse.py
@@ -38,13 +38,20 @@ def parse_irr(context):
 
         # Parse the RPSL objects in the IRR DB into Python Dicts
         for line in lines:
-            if line == '\n':
-                entry_list.append(current_entry)
-                current_entry = {}
-            else:
-                if ":" in line:
-                    k, v = line.strip().split(':', 1)
-                    current_entry[k.strip()] = v.strip()
+            if not line.strip():
+                if current_entry:
+                    entry_list.append(current_entry)
+                    current_entry = {}
+                continue
+
+            if ":" in line:
+                k, v = line.strip().split(':', 1)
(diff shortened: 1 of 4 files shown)


fjahr:
Please try to explain what the problem(s) are that you are solving in the PR rather than describing only the change. You should also make a separate commit for each fix, ideally with a test that did not pass before the fix but passes afterwards.

> Validation:

Please also test consistency with a full reproduction run from `asmap-data`, thanks!

jorisstrakeljahn:
I reviewed this against current master on the same kind of reproduce inputs we use for collaborative launches.

The IRR parse half is a real bug: `parse_irr` never flushes the last RPSL object when the dump has no trailing blank, so that object is dropped. Write-up, full `-irr -rv` rebuild numbers on `1780588800` / `1783008000`, and a focused fix are in PR #148.

I would not land the rest of this PR as-is:

1. The merge rewrite (how the output file is written). On current master, running the same merge twice already gives the same bytes, and I did not see blank-line junk in real run outputs. So this does not look like a correctness fix.
2. Capping merge workers at 4. That only changes how many processes run in parallel. The merge result stayed the same when I tried 4 vs the default on this machine.

If that sounds right, this PR could be closed and the parse fix taken via #148. Thanks @SuvidhJ for raising the edge case here and in #138.

cc @fjahr

PR #139, opened 2026-04-21
warn: include detected rpki-client version in compatibility warning

brunoerg:
Just a minor change to tell user which version of rpki-client is running when warning about untested versions.

### DIFF
--- a/kartograf/util.py
+++ b/kartograf/util.py
@@ -90,7 +90,8 @@ def check_compatibility():
     elif local_version > latest_version:
         print(
             "Warning: This kartograf version has not been tested with "
-            f"rpki-client versions higher than {latest_version}."
+            f"rpki-client versions higher than {latest_version}. "
+            f"You are running rpki-client {local_version}."
         )
     else:
         print(


fjahr:
Oh, right, this is the only path that isn't printing the used version, good catch!

kartograf/util.py
```diff
@@ -89,8 +89,9 @@ def check_compatibility():
         print(f"Using rpki-client version {local_version} (recommended).")
     elif local_version > latest_version:
         print(
-            "Warning: This kartograf version has not been tested with "
-            f"rpki-client versions higher than {latest_version}."
+            f"Warning: This kartograf version has not been tested with "
```
fjahr: You added the `f` here but there is not interpolation on this line and the line is otherwise unchanged. A bit surprised that the linter doesn't complain about this TBH :) Could you remove it again?

brunoerg: Oh, true, just removed it. Thanks.

fjahr:
ACK fe7a0332d73501b3963bb5ebe9f42317c305901c

PR #140, opened 2026-04-21
Resolve RouteViews URLs early

fjahr:
This contains the first two commits of #129, since the rest of the change needs more work but we would like to get this fix in before the next collaborative run and the original author hasn't responded to the latest feedback yet.


fjahr:
These commits are unchanged from the original PR and received enough review there. I have tested them multiple times so I think this is ready to merge.

PR #143, opened 2026-06-03
use a tempdir instead of the rpki output dir for rpki-client output

jurraca:
a by-product of aee081eea07a186657b47e6eb086bc9178cf1bec was the introduction of a writable directory explicitly in the rpki-client run command invocation, which now gets all rpki-client output written to it. This directory is intended for kartograf output, not rpki-client. It seems that systems with writable directories (non-Nix) write this data to a default directory managed by rpki-client, and Nix systems don't write the data at all, failing silently. 

The better solution is to name an explicit, temporary directory for rpki-client output and pass it to the `run` command. The temporary directory is cleaned up automatically after the run finishes.

The `run` command defaults to writing `-joBcm` outputs, but we don't need the output. We add the `-m` flag to the run args, since it limits the output written to the smallest item (metrics, ~330K). 

To summarize: we set an explicit, temporary output directory, we write a tiny bit of data to it, and it gets cleaned up automatically.

### DIFF
--- a/kartograf/rpki/fetch.py
+++ b/kartograf/rpki/fetch.py
@@ -1,6 +1,7 @@
 import re
 import subprocess
 import sys
+from tempfile import TemporaryDirectory
 
 from concurrent.futures import ThreadPoolExecutor, as_completed
 import json
@@ -73,35 +74,32 @@ def data_tals(context):
 def fetch_rpki_db(context):
     # Download TALs and presist them in the RPKI data folder
     download_rir_tals(context)
-    tal_options = [item for path in data_tals(context) for item in ('-t', path)]
-    run_args = ["rpki-client", "-d", context.data_dir_rpki_cache,
-                 "-P", context.epoch] + tal_options
-    print("Downloading RPKI Data, this may take a while.")
+    with TemporaryDirectory(prefix="rpki-out-") as tmpdir:
+        tal_options = [item for path in data_tals(context) for item in ('-t', path)]
+        run_args = ["rpki-client", "-m", "-d", context.data_dir_rpki_cache,
+                     "-P", context.epoch] + tal_options
+        print("Downloading RPKI Data, this may take a while.")
 
-    if context.stable_repos:
-        for url in STABLE_REPO_URLS:
-            run_args += ["-H", url]
-        print("Using only stable RPKI repositories.")
+        if context.stable_repos:
+            for url in STABLE_REPO_URLS:
+                run_args += ["-H", url]
+            print("Using only stable RPKI repositories.")
 
-    # rpki-client requires a writable output directory as a positional
-    # argument. Without one it falls back to a compiled-in default that
-    # is not writable in some environments (e.g. nix), which makes it
-    # exit before emitting the CCR hashes.
-    run_args.append(context.out_dir_rpki)
+        run_args += [tmpdir]
 
-    result = subprocess.run(run_args,
-                            capture_output=True,
-                            check=False)
+        result = subprocess.run(run_args,
+                                capture_output=True,
+                                check=False)
 
-    if context.debug_log:
-        with open(context.debug_log, 'a') as logs:
-            logs.write("=== RPKI Download ===\n")
-            if result.stdout:
-                logs.write(result.stdout.decode())
-            if result.stderr:
-                logs.write(result.stderr.decode())
+        if context.debug_log:
+            with open(context.debug_log, 'a') as logs:
+                logs.write("=== RPKI Download ===\n")
+                if result.stdout:
+                    logs.write(result.stdout.decode())
+                if result.stderr:
+                    logs.write(result.stderr.decode())
 
-    parse_ccr_hashes(result.stdout.decode() if result.stdout else "")
+        parse_ccr_hashes(result.stdout.decode() if result.stdout else "")
 
     print(f"Downloaded RPKI Data, hash sum: {calculate_sha256_directory(context.data_dir_rpki_cache)}")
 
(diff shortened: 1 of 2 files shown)


fjahr:
This breaks the stable repos flag because, if used, the flag gets appended after the temp dir and then it is ignored (see the download time in the logs, it doesn't download anything, thats why it is so fast).

```
$ ./run map -irr -rv -s

--- Start Kartograf ---

Kartograf version: 0.4.13
Using rpki-client version 9.8 (recommended).
The epoch for this run is: 1780589799 (2026-06-04 16:16:39 UTC, local: 2026-06-04 18:16:39 CEST)

--- Resolving RouteViews URLs ---

Resolved RouteViews URLs:
  v4: https://publicdata.caida.org/datasets/routing/routeviews-prefix2as/2026/06/routeviews-rv2-20260603-1200.pfx2as.gz
  v6: https://publicdata.caida.org/datasets/routing/routeviews6-prefix2as/2026/06/routeviews-rv6-20260603-1200.pfx2as.gz

--- Fetching RPKI ---

Downloaded TAL for AFRINIC to afrinic.tal, file hash: 2838ef30ea27ce5705abf5f5adb131d8c35b1f50858338a2f3c84bb207c2fa35
Downloaded TAL for APNIC to apnic.tal, file hash: 472e551f7c551c2e999e582b7c9437d3bee4900fe53afff62aeb28d4940ade94
Downloaded TAL for ARIN to arin.tal, file hash: 1f8bdb03bcc30a3b8e11fd9a87102fba250c22137a3c8baa9c81b139cb412639
Downloaded TAL for LACNIC to lacnic.tal, file hash: d44bb9394ab009c8b53e5efebf2a1c9450bab61a27efe00de5a3e4587a3a2f6a
Downloaded TAL for RIPE to ripe.tal, file hash: 59ca27ef93f23682749fcefe7c6d70fbc723343549ff9e4d3996acaff79817fb
Downloading RPKI Data, this may take a while.
Using only stable RPKI repositories.
Downloaded RPKI Data, hash sum: e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855
...finished in 0:00:05.356664
```

jurraca:
thanks, should have caught that. fixed with [2a983f2](https://github.com/asmap/kartograf/pull/143/commits/2a983f2f8d939d9bf725cbff9eebce8732c8f647) by appending the `tmpdir` after the `-s` appending clause.

kartograf/rpki/fetch.py
```diff
@@ -73,35 +74,32 @@ def data_tals(context):
...
-
-    if context.stable_repos:
-        for url in STABLE_REPO_URLS:
-            run_args += ["-H", url]
-        print("Using only stable RPKI repositories.")
-
-    # rpki-client requires a writable output directory as a positional
-    # argument. Without one it falls back to a compiled-in default that
-    # is not writable in some environments (e.g. nix), which makes it
-    # exit before emitting the CCR hashes.
```
fjahr, line 89: Why is this comment removed? It's still valid and valuable information IMO. Maybe also add a similar comment in parse. And I think this could be even more strongly worded that there is nothing written to the folder that we care about, which is why it's ok to use the temp folder here.

kartograf/rpki/fetch.py
```diff
@@ -73,35 +74,32 @@ def data_tals(context):
...
-            logs.write("=== RPKI Download ===\n")
-            if result.stdout:
-                logs.write(result.stdout.decode())
-            if result.stderr:
-                logs.write(result.stderr.decode())
-
-    parse_ccr_hashes(result.stdout.decode() if result.stdout else "")
+    with TemporaryDirectory(prefix="rpki-out-") as tmpdir:
+        tal_options = [item for path in data_tals(context) for item in ('-t', path)]
+        run_args = ["rpki-client", "-m", "-d", context.data_dir_rpki_cache,
```
fjahr, line 90: Not sure if we really need the `-m` as optimization since we are deleting the files anyway. If you want to keep it, please add a small comment why it's there (also below). For the other parameters it seems intuitive why they are needed but for `-m` I'm not sure I will remember why it's there in a year ;)

fjahr (commented):
Logic looks good but I would like to see some more documentation here because it doesn't seem intuitive at all why this is being done the way it is. I will do some further manual testing. Then, with the added docs, should be good to go!

kartograf/rpki/fetch.py
jurraca: fair, brought it back with [1bf32c2](https://github.com/asmap/kartograf/pull/143/commits/1bf32c24b538cf05e2b60771e4839cf3a1a510c5). I duplicated the docs in rpki/fetch.py and rpki/parse.py.

kartograf/rpki/fetch.py
jurraca: it just reduces the amount of data we write before deleting, which is an optimization. there's no need to write data we won't need, so it makes sense imo.

kartograf/rpki/fetch.py
```diff
@@ -71,37 +72,45 @@ def data_tals(context):
...
+    argument. Without one it falls back to a compiled-in default that
+    is not writable in some environments (e.g. nix).
+    We use python's tempdir, which gets cleaned up automatically.
+    All kartograf output we care about is written to the context.out_dir_rpki.
+
+    We use the '-m' flag to only output metrics to this temporary directory,
+    since the default writes several hundred MB of data. The metrics
+    data is not used.
+    """
+    # Download TALs and persist them in the RPKI data folder
```
fjahr, line 86: nit: This should have been part of the docstring

fjahr (commented):
tACK 1bf32c24b538cf05e2b60771e4839cf3a1a510c5

PR #145, opened 2026-06-30
check prefixlen in check_inclusion

### LINKED ISSUE #144: Merging does not add networks if a subnet exists
While investigating #123 , I found a bug in our merge process. While the current and trie process should be identical, certain network assignments were not getting merged in as expected, causing a diff between output files. Note that this is a different scenario than the original bug report, which was incorrect. The problem is our `check_inclusion` function.

```
    def check_inclusion(self, row, root_net, version):
        """
        A network is a subnet of another if the bitwise AND of its IP and the base network's netmask
        is equal to the base network IP.
        """
        for net, mask in self._dict[version][root_net]:
            if row.INETS & mask == net:
                return 1
        return 0
```
Given two networks, `12.0.0.0/8` and a more specific subnet like `12.0.0.0/22`, if the more specific one exists in the base file, and the extra file contains the less specific one (`/8`), the check will return `True`. The merge will reject the less specific network, but it is _not_ included in the more specific one, even though they overlap. 

simulating our check_inclusion logic:
```
Python 3.13.11 (main, Dec  5 2025, 16:06:33) [GCC 14.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import ipaddress
>>> ip1 = "12.0.0.0/8"
>>> ip2 = "12.0.0.0/22"
>>> netw1 = int(ipaddress.ip_network(ip1).network_address)
>>> netw2 = int(ipaddress.ip_network(ip2).network_address)
>>> mask2 = int(ipaddress.ip_network(ip2).netmask)
# inclusion check
>>> netw1 & mask2 == netw2
True
```
It is true, but we want this to be false, i.e. `/8` is not included in `/22`. A test [here](https://github.com/jurraca/kartograf/commit/5e112a6572fe245e313eda5737529a1cc3628186) reproduces the failure.  

The solution is to include a prefixlen check when checking inclusion, or use `subnet_of()`. We would have to add the prefixlen to the map we store and only check inclusion if the prefixlen is larger than the existing one, i.e.: `if (row.INETS & mask == net) & (row.prefixlen >= net_prefixlen)`. 

The trie implementation does not have this issue, so fixing the current impl is up for discussion.

jurraca:
When merging in a map (such as IRR into RPKI), a more specific prefix should not supersede a less specific network. This adds a check on the prefix length of the network, which returns check_inclusion = True only if the networks match and the prefix length is greater (edit: or equal to) ~than~ the existing one.

The diff from the [1780588800](https://github.com/bitcoin-core/asmap-data/issues/51) run with the same data sources yields 16856 lines, the majority of which are un-assignments, i.e. many networks that were previously merged in due to the current check_inclusion are no longer assigned, as expected.

This is backwards incompatible, in the sense that given a same input, the output ASmap will be different if this is merged in.

Closes #144

### DIFF
--- a/kartograf/merge.py
+++ b/kartograf/merge.py
@@ -41,17 +41,19 @@ class BaseNetworkIndex:
 
         if (root_net in self._v4_keys) or (root_net in self._v6_keys):
             current = self._dict[v][root_net]
-            self._dict[v][root_net] = current + [(netw, mask)]
+            self._dict[v][root_net] = current + [(netw, mask, ipn.prefixlen)]
         else:
-            self._dict[v].update({root_net: [(netw, mask)]})
+            self._dict[v].update({root_net: [(netw, mask, ipn.prefixlen)]})
 
     def check_inclusion(self, row, root_net, version):
         """
         A network is a subnet of another if the bitwise AND of its IP and the base network's netmask
-        is equal to the base network IP.
+        is equal to the base network IP, and the network's prefix length is larger or equal than the base network's
+        prefix length.
         """
-        for net, mask in self._dict[version][root_net]:
-            if row.INETS & mask == net:
+        candidate = ipaddress.ip_network(row.PFXS)
+        for net, mask, prefixlen in self._dict[version][root_net]:
+            if candidate.prefixlen >= prefixlen and row.INETS & mask == net:
                 return 1
         return 0
 
--- a/tests/test_merge.py
+++ b/tests/test_merge.py
@@ -126,6 +126,41 @@ def test_merge_rpki_supersedes_irr_subnets(tmp_path):
         f"IRR subnets should not appear in merged output: {set(final_ips) & set(irr_ips)}"
     )
 
+def test_more_specific_base_does_not_suppress_extra_supernet(tmp_path):
+    '''
+    Test that a base more-specific prefix does not suppress an extra supernet
+    with the same network address.
+    A more specific network should not supersede a less specific network merged first.
+    In this test, 12.2.12.0/24 should not be merged since it is more specific than
+    12.0.0.0/8 and /9.
+    '''
+    rpki_path = tmp_path / "rpki_final.txt"
+    irr_path = tmp_path / "irr_final.txt"
+    rv_path = tmp_path / "pfx2asn_clean.txt"
+    merged_rpki_irr_path = tmp_path / "merged_file_rpki_irr.txt"
+    final_path = tmp_path / "merged_file_rpki_irr_rv.txt"
+
+    rpki_path.write_text("12.0.0.0/22 AS6269\n")
+    irr_path.write_text("12.0.0.0/8 AS7018\n12.0.0.0/9 AS7018\n")
+    rv_path.write_text("12.2.12.0/24 AS40724\n13.0.0.0/8 AS1239\n")
+
+    general_merge(rpki_path, irr_path, None, merged_rpki_irr_path)
+    general_merge(merged_rpki_irr_path, rv_path, None, final_path)
+
+    merged_rpki_irr = merged_rpki_irr_path.read_text()
+    final_result = final_path.read_text()
+
+    assert "12.0.0.0/22 AS6269\n" in merged_rpki_irr
+    assert "12.0.0.0/8 AS7018\n" in merged_rpki_irr
+    assert "12.0.0.0/9 AS7018\n" in merged_rpki_irr
+
+    assert "12.0.0.0/22 AS6269\n" in final_result
+    assert "12.0.0.0/8 AS7018\n" in final_result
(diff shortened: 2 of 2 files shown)


fjahr:
> many networks that were previously merged in due to the current check_inclusion are no longer assigned, as expected

Huh, this confuses me. I thought we would be getting more assignments with this since we are accepting assignments that were previously rejected?

kartograf/merge.py
```diff
@@ -41,17 +41,18 @@ def update(self, pfx):
...
-            self._dict[v][root_net] = current + [(netw, mask)]
+            self._dict[v][root_net] = current + [(netw, mask, ipn.prefixlen)]
         else:
-            self._dict[v].update({root_net: [(netw, mask)]})
+            self._dict[v].update({root_net: [(netw, mask, ipn.prefixlen)]})
 
     def check_inclusion(self, row, root_net, version):
         """
         A network is a subnet of another if the bitwise AND of its IP and the base network's netmask
         is equal to the base network IP.
```
fjahr: Documentation of the function should be updated as well

fjahr:
> This is backwards incompatible

We said we don't really do versions anymore but we still have them so I would say we should bump it to v0.5.0 with this change.

fjahr:
>  only if the networks match and the prefix length is greater than the existing one

Description nit: It's "greater than or equal", right?

fjahr:
Concept ACK obviously, thanks for extracting this separate PR.

jurraca:
> I thought we would be getting more assignments with this since we are accepting assignments that were previously rejected?

Sorry, yes, the diffs shows mostly previously unassigned networks (16076 out of 16856), so there's more assignments in the new file. Though it may not always be the case, depending on what IRR and RV provide. For example, in the test, new IRR networks are added but the more specific RV networks are rejected, even though they would be included previously. diff file: [fix-diff.txt](https://github.com/user-attachments/files/29551242/fix-diff.txt)

jurraca:
> We said we don't really do versions anymore but we still have them so I would say we should bump it to v0.5.0 with this change.

ACK on 0.5.0, makes sense. imo we should still do versions in general, any reason not to?

jorisstrakeljahn:
I ran the historical analysis discussed in #144: all four collaborative launches since the embedded map (`1772726400` through `1783008000`), rebuilt from the official asmap-data raw inputs, once on master before this PR (2249ea8) and once with it (8e70dd2). Full write-up with method, tables and repro steps: [Historical impact of kartograf#145](https://gist.github.com/jorisstrakeljahn/30235e943db90f78fecb9e397ffee9c5)

Summary:

| Launch | Diff lines | New assignments | Reassignments | Unassignments | Affected clearnet peers |
|---|---|---|---|---|---|
| `1772726400` (embedded) | 16,260 | 15,521 | 739 | 0 | 3 newly mapped |
| `1776960000` | 16,403 | 15,677 | 726 | 0 | 4 newly mapped |
| `1780588800` | 16,852 | 16,076 | 776 | 0 | 3 newly mapped |
| `1783008000` | 17,379 | 16,690 | 689 | 0 | 4 newly mapped |

- The builds without the fix reproduced the attested `final_result.txt` hashes on all four launches, so the baseline is exactly what shipped.
- Nearly all diff lines are prefixes gaining an assignment, and no prefix loses one. On `1780588800` my line count matches the one in the PR description (16,852 against 16,856). Comparing the old map to the new map, these lines are new assignments, which is also what the fix predicts, since the old check dropped covering routes.
- The restored coverage is concentrated in large carriers. On the embedded map, about 40 percent of the newly covered IPv4 space belongs to AS7018 (AT&T), whose covering IRR routes for `12.0.0.0/8` the old check dropped.
- One side effect: once a large IRR supernet is admitted, more specific RouteViews prefixes with a different ASN inside it are skipped, so some lookups now return the provider ASN instead of a customer ASN. On the embedded map, 484 of the 739 reassignments end at AS7018. No peer in the snapshots was affected by this, but it is a granularity tradeoff to keep in mind.
- Node impact is small. Peer sets are bitnodes.io exports from the launch days (March, April) and bitnod.es exports at or near the launch days (June, July), clearnet only. Across all four launches, five distinct peers move from unmapped to mapped (three in AT&T space, two now AS5056), and no peer moved from one AS to another.

I think this can merge. The coarsening side effect could be a follow-up discussion (an ASN-aware merge), but it is not a reason to keep the bug.

jurraca:
Thanks for running the numbers @jorisstrakeljahn ! 
From my perspective, this change makes sense and its impact is minimal. While it does affect our reproducibility (as noted by the CI), it makes the map generation more correct.

fjahr:
ACK 8e70dd2647f88ea10e7c6930c78ca6f953d58d16

PR #146, opened 2026-07-02
Fail early if RPKI fetch and parse don't output assignments

jurraca:
RPKI can sometimes fail to write its cache data to disk, or return no valid assignments after validation. In both cases, we want to exit early, since RPKI is the basis upon which we try to build a more complete map. 

For the parsing test, I modified the incompletes test as well to specifically test the incompletes output with at least one valid entry.

### DIFF
--- a/kartograf/rpki/fetch.py
+++ b/kartograf/rpki/fetch.py
@@ -121,10 +121,11 @@ def validate_rpki_db(context):
              if path.is_file() and ((path.suffix == ".roa")
                                     or (path.name == ".roa"))]
 
-    print(f"{len(files)} raw RKPI ROA files found.")
     if len(files) == 0:
         print("No RPKI files found! Exiting.")
         sys.exit(1)
+    else:
+        print(f"{len(files)} raw RPKI ROA files found.")
 
     rpki_raw_file = 'rpki_raw.json'
     result_path = Path(context.out_dir_rpki) / rpki_raw_file
--- a/kartograf/rpki/fetch.py
+++ b/kartograf/rpki/fetch.py
@@ -122,6 +122,10 @@ def validate_rpki_db(context):
                                     or (path.name == ".roa"))]
 
     print(f"{len(files)} raw RKPI ROA files found.")
+    if len(files) == 0:
+        print("No RPKI files found! Exiting.")
+        sys.exit(1)
+
     rpki_raw_file = 'rpki_raw.json'
     result_path = Path(context.out_dir_rpki) / rpki_raw_file
 
--- a/kartograf/rpki/parse.py
+++ b/kartograf/rpki/parse.py
@@ -1,6 +1,7 @@
 import json
 import subprocess
 from pathlib import Path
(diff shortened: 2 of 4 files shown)


jorisstrakeljahn:
Concept ACK. I opened #147 as a possible follow-up to this (exit codes, stderr reasons, per-RIR counts), they should work independently.

kartograf/rpki/fetch.py
```diff
@@ -122,6 +122,10 @@ def validate_rpki_db(context):
                                     or (path.name == ".roa"))]
 
     print(f"{len(files)} raw RKPI ROA files found.")
+    if len(files) == 0:
```
fjahr, line 125: It doesn't really make sense to print the line above in this case, so I would move this above it.

fjahr: There is also a typo in the log above that you can fix while you are at it: "RKPI"

fjahr (commented):
I am ok to add this but it would be much more interesting to figure out what the errors really are that we are running into and try to catch these before the actual run, like in our pre-checks ideally. 

For a OOM scenario I think the most likely place for this to be caught should be here: https://github.com/asmap/kartograf/pull/146/files#diff-5d8704566493265b100a183244c9429c1424306cb6f7d14ab32e76b19018e4acR173 If a batch doesn't return a result then it was likely killed. Maybe this could be addressed here as well.

But for the empty files case I don't think it makes sense that this could be an OOM case, there is no memory hungry operation running at that point. It's just rpki-client downloading stuff. So I am still wondering what the real cause there is.

Can you open a tracking issue for these cases so we can collect them in one place and try to coordinate reproducers?

fjahr:
> For a OOM scenario I think the most likely place for this to be caught should be here

Looks like @jorisstrakeljahn is looking into this in #147 already

kartograf/rpki/parse.py
```diff
@@ -114,6 +115,10 @@ def parse_rpki(context):
 
     context.cleanup_out_files.append(raw_input)
 
+    if out_count == 0:
```
fjahr: I think this can be actually moved below the 5 reporting prints below. These may still have some valuable debugging information after all.

fjahr (commented):
If you could address my review comments (the old ones and the one new one) I think this is ready for merge. Maybe do a rebase as well for good measure.

kartograf/rpki/fetch.py
jurraca: does this work better as an if/else ?

jurraca:
thanks for the re-review, forgot about this PR! updated and rebased.

kartograf/rpki/fetch.py
fjahr: I prefer without `else` since it's redundant if the `if` exits/is a guard condition, but not that important.

fjahr:
ACK 66dd706f9f5463e6a60cc0c761fb538a5821141d

PR #147, opened 2026-07-03
Fail on rpki-client errors and surface validation problems

jorisstrakeljahn:
Follow-up to the discussion in bitcoin-core/asmap-data#60. 
Kartograf currently ignores rpki-client's exit code and drops its stderr unless --debug is set. All the failures we saw degraded silently into plausible looking but wrong maps: an empty cache (69MB result), an all-invalid validation (sipa's 72MB result) and a partial fetch (nkaretnikov's 62MB result in bitcoin-core/asmap-data#51).

This fails when rpki-client exits with an error, when validation results get lost (e.g. a batch got killed) and when the majority of ROAs is invalid. 

A reproduction run over the 1783008000 data with these changes produces the identical result hash (35a4b28f), so the checks are invisible on healthy runs. Complements #146.

### DIFF
--- a/kartograf/rpki/fetch.py
+++ b/kartograf/rpki/fetch.py
@@ -110,6 +110,13 @@ def fetch_rpki_db(context):
                 if result.stderr:
                     logs.write(result.stderr.decode())
 
+        if result.returncode != 0:
+            print(f"rpki-client exited with code {result.returncode} "
+                  "during fetch:")
+            if result.stderr:
+                print("\n".join(result.stderr.decode().splitlines()[-10:]))
+            sys.exit(1)
+
         parse_ccr_hashes(result.stdout.decode() if result.stdout else "")
 
     print(f"Downloaded RPKI Data, hash sum: {calculate_sha256_directory(context.data_dir_rpki_cache)}")
--- a/kartograf/rpki/fetch.py
+++ b/kartograf/rpki/fetch.py
@@ -158,6 +158,12 @@ def validate_rpki_db(context):
                                  capture_output=True,
                                  check=False)
 
+        if result.returncode != 0:
+            print("rpki-client validation batch failed "
+                  f"(exit {result.returncode}):")
+            print(result.stderr.decode() if result.stderr else "(no stderr)")
+            sys.exit(1)
+
         if result.stderr and context.debug_log:
             stderr_output = result.stderr.decode()
             with debug_file_lock:
@@ -186,6 +192,12 @@ def validate_rpki_db(context):
 
         with open(result_path, 'w') as f:
             json.dump(s, f)
+
+    # rpki-client emits exactly one JSON object per -f input file
+    if len(results_json) != len(files):
+        print(f"Validation incomplete: {len(results_json)} results for "
+              f"{len(files)} files! Exiting.")
+        sys.exit(1)
 
     print(f"{len(results_json)} RKPI ROAs validated\nSaved to: {result_path.name}\nFile hash: {calculate_sha256(result_path)}")
 


kartograf/rpki/parse.py
```diff
@@ -114,6 +115,14 @@ def parse_rpki(context):
 
     context.cleanup_out_files.append(raw_input)
 
+    # Healthy runs see well under 1% invalid ROAs; a failing majority
+    # means the local trust chain is broken, not the data.
+    valid_count = len(data) - invalids - incompletes - not_roas
```
fjahr, line 120: I think this check seems pretty arbitrary, did we see any actual cases where this would have triggered?

kartograf/rpki/fetch.py
```diff
@@ -215,3 +222,22 @@ def parse_ccr_hashes(output):
...
+    comparing logs. Warns instead of aborting because a RIR outage can
+    be global, affecting all participants equally.
+    """
+    pattern = re.compile(
+        r'rpki_client_ta_objects\{type="roa",state="valid",name="(\w+)"\} (\d+)')
+    counts = pattern.findall(metrics_path.read_text())
+    if not counts:
+        return
+    print("Valid ROAs per RIR: "
+          + ", ".join(f"{tal}: {n}" for tal, n in counts))
```
fjahr: `tal` is not the right variable name here, it's just the name of an RIR

kartograf/rpki/fetch.py
```diff
@@ -120,6 +121,12 @@ def fetch_rpki_db(context):
 
         parse_ccr_hashes(result.stdout.decode() if result.stdout else "")
 
+        metrics_file = Path(tmpdir) / "metrics"
```
fjahr: This file path handling doesn't belong here. We write the metrics file to tmpdir for a reason, you shouldn't copy it back here. If we think it has valuable info we shouldn't write it to temp to begin with. This layer should be generally handled in the context file.

I think this metric is nice to have but I think it doesn't help with debugging. The more realistic failure case is that a certain repository was not available but that doesn't mean that any RIR has no entries here then because repositories can have ROAs of different RIRs. This is data that could be in debugging but really I think it's something that can be analyzed after the fact if there are large mismatches.

I also don't know much about this metrics file other than that it was just very recently introduced. Not sure if we should rely on it generally. Overall I tend to think this should be dropped.

kartograf/rpki/fetch.py
```diff
@@ -215,3 +222,22 @@ def parse_ccr_hashes(output):
...
+
+
+def report_tal_roa_counts(metrics_path):
+    """
+    Print valid ROAs per RIR so participants can spot missing RIRs when
+    comparing logs. Warns instead of aborting because a RIR outage can
+    be global, affecting all participants equally.
+    """
+    pattern = re.compile(
+        r'rpki_client_ta_objects\{type="roa",state="valid",name="(\w+)"\} (\d+)')
```
fjahr: It's interesting that rpki-client does the validation right away with the download step but we can't really rely on this because this validation is not fixed to timestamp of the run like we do for the actual validation later. So there will likely be minor differences between the counts which makes them less interesting to show IMO. It's also confusing to the readers of the logs because we would be reporting counts for valid ROAs only to the do the actual RPKI validation step afterwards.

kartograf/rpki/fetch.py
```diff
@@ -182,6 +194,18 @@ def process_files_batch(batch):
...
             json.dump(s, f)
 
+    # rpki-client emits exactly one JSON object per -f input file
+    if len(results_json) != len(files):
+        print(f"Validation incomplete: {len(results_json)} results for "
+              f"{len(files)} files! Exiting.")
+        sys.exit(1)
+
+    if error_counter:
+        print(f"Validation reported {sum(error_counter.values())} problems, "
```
fjahr: I don't think this is valuable. It is known that `rpki-client` reports lots of errors on every run: https://console.rpki-client.org/ If we think that this is valuable information we allow users to run with debug and then errors can be inspected directly.

kartograf/rpki/fetch.py
```diff
@@ -153,11 +155,21 @@ def process_files_batch(batch):
                                  capture_output=True,
                                  check=False)
 
-        if result.stderr and context.debug_log:
+        if result.returncode != 0:
+            print("rpki-client validation batch failed "
```
fjahr, line 157: This is different from the error counting and should be in a separate commit. This is worth keeping.

fjahr (commented):
Concept ACK for the batch failures/return code stuff.

The verbose logging stuff isn't really adding value IMO. We spent quite a bit of time trimming the default logs and these don't seem to be providing enough value to make the cut.

In general, when we throw away stuff like the metrics file or we use stuff only in the debug context currently, it would be better to first ask why that is. If there is a good reason that the info there is more valuable than we initially thought, then these files should be first made not throwaway/debug only anymore before using their contents in the default logs.

kartograf/rpki/parse.py
jorisstrakeljahn: It would have triggered for sipa's run. All 357335 ROAs failed validation there (bitcoin-core/asmap-data#60). To be fair, with #146 merged that exact case is also caught by the out_count check, so the extra coverage here is partial failures. On the cutoff: healthy runs stay well under 1% invalids (0.66% was the highest I found), sipa's was 100%, so majority is just the most conservative line between the two, not a tuned number. Happy to drop it if you think #146 covers enough.

kartograf/rpki/fetch.py
jorisstrakeljahn: Fair points, dropped the commit. If the per-RIR counts ever turn out useful for comparing diverging runs, I agree that should go through the context layer properly instead.

kartograf/rpki/fetch.py
jorisstrakeljahn: Agreed that these are mostly noise on healthy runs, but my worry was that in sipa's case nobody ran with --debug so the reason was lost, but I see the tradeoff with keeping the default logs clean. Dropped that

kartograf/rpki/fetch.py
jorisstrakeljahn: Done, it's its own commit now

kartograf/rpki/fetch.py
jorisstrakeljahn: Dropped the per-RIR reporting entirely, see the other threads

kartograf/rpki/fetch.py
jorisstrakeljahn: Dropped anyway, but one small note: 
The fetch validation is also pinned to the epoch (kartograf passes -P there as well), so the small count differences come from the -f mode revalidation each file standalone, not from timestamps. On my run it was 11 failed at fetch vs 12 invalid in the -f validation, out of 355.016 over the same cache.  And a reproduction run two weeks later still gives exactly the same 12, which wouldn't happen if the wall clock was involved i guess.

jorisstrakeljahn:
Thanks for the review and sorry for the delay.
Dropped the metrics/per-RIR commit (was b2ba497) and removed the stderr summary, so the batch failure checks is its own commit now as suggested. Kept the majority check for now, see the thread there.

kartograf/rpki/parse.py
fjahr: I still think this should be dropped. It's just too arbitrary for my taste and as you said, the actual cases we saw are already covered with other checks.

I think something that we could still do to catch the 62 MB case in https://github.com/bitcoin-core/asmap-data/issues/51 could be to check if we got some data from each of the repositories we are using, at least in the stable repositories case. That should be doable without using the metrics file, just looking at the folders. But still unsure if that would really be worth it and work well enough. If you want to try this out please open a new pull for this.

fjahr:
Aside from dropping the third commit this is ready for merge.

fjahr:
And please rebase for good measure :)

fjahr:
ACK 8821772b9d2f948934a6d17d0d0d9dab0e63914d

I dropped the third commit here since @jorisstrakeljahn isn't available right now. I will merge when CI is green.

PR #153, opened 2026-08-29
Remove dependencies bs4, pandas, tqdm

fjahr:
Experimental re-implementation of the features we needed, which was possible with far less code than I thought. The tqdm progress bar became a lot less relevant since we are running a lot faster now anyway and that should already be the most notable change for users.

I reproduced the list of hashes from https://github.com/asmap/kartograf/pull/151 with this, so there there is good evidence this is consistent. It seems to be also slightly faster but it still requires more testing.


jurraca (commented):
interesting, but this froze my machine for a good 40 minutes on the merge step before crashing :) 

the bs4 -> html.parser and tqdm changes seem harmless, but using pandas has some significant memory benefits compared to pure lists, so I think this would hurt lower memory devices. This mostly goes away with the trie implementation though.

kartograf/merge.py
```diff
@@ -27,19 +26,8 @@ def update(self, pfx, asn):
...
-        try:
-            candidate = ipaddress.ip_network(row.PFXS)
-        except ValueError:
-            return 0
-        asn = self._trie.covering_asn(candidate)
-        if asn is not None:
-            return 1
-        return 0
+    def contains(self, pfx):
+        return self._trie.covering_asn(ipaddress.ip_network(pfx)) is not None
```
jurraca, line 30: nice simplification :)

jurraca (approved):
tACK https://github.com/asmap/kartograf/commit/4e43e23e5831f438ff947007dfc2347e78870a3a 

Memory looks good now.

PR #155, opened 2026-08-29
tests: Skip generated subnets that already exist in the base networks

fjahr:
This fixes a flaky test issue that was observed in #123 and I described what I think the issue is here: https://github.com/asmap/kartograf/pull/123#issuecomment-5309879070


jurraca (approved):
nice, thanks

PR #157, opened 2026-09-04
Prune entries that don't change the encoded asmap file

fjahr:
Our results include redundant entries, e.g. longer prefixes with the same ASN as a shorter prefix that includes this prefix. This was primarily due to debugging purposes and we left the pruning of redundant information to the encoding step.

We never saw any issues coming from this in terms of result matches in collaborative runs until yesterday, this caused the confusing result here: https://github.com/bitcoin-core/asmap-data/issues/68#issuecomment-5531088208 There is a mismatch in the hash despite the coverage being the same (this is what the karto-server diff shows).

With these changes we discard redudant entries in each result files and then once more at the end. Doing it only at the end would be simpler but since the merge step it the most expensive part of our processing we can see some nice speed-ups from doing it for each source before merging because there is simply fewer data to merge.

While this makes a significant difference for the file coming out of the kartograf step the results are the same for the encoded result. I verified this for the last three runs. Thus I wouldn't call this a breaking change in the same way as our recent changes in `0.5.0`. But our reproduction CI job needs to be updated.

The logging of the pruned entries is a bit unnecesary but at least for the start I would like to have a closer eye on what is happening with this change and I hope we can remove this in a few weeks when we don't see any issues.

### DIFF
--- a/kartograf/sort.py
+++ b/kartograf/sort.py
@@ -1,6 +1,7 @@
 import ipaddress
 from pathlib import Path
 
+from kartograf.prune import prune_entries
 from kartograf.timed import timed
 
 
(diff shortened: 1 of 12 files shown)


jurraca (approved):
ACK b668f794a6bdb866439344895757869c819d4af7 

pretty wild. I didn't realize there were so many redundant entries in the input files. The performance gain is very welcome. Doing a reproduction run on my 1788451200 data, the final_result file goes from 36M to 17M (!). 

`prune_entries` is efficient and test coverage looks good.

kartograf/sort.py
jorisstrakeljahn: I noticed the reconstruct after the sort in sort_result_by_pfx looks leftover. prune_entries already returns canonical prefix strings, so the longer-first sort is still needed (prune is shorter-first), but after that it can write the original strings instead of rebuilding IPv4Network/IPv6Network I guess.

Not blocking and doesn't have to go in this PR. @fjahr I can open a follow-up if you want to keep this one focused, because I think this ist also not time critical.

fjahr: Sounds good, I pushed this as a small refactor, will merge when CI is green since it isn't a big change and the core of the change was not touched.

PR #162, opened 2026-09-08
Add log debug helper, fix bogon logging

jurraca:
When parsing invalid prefixes and ASNs, and if debug is enabled, we write lines to `debug.log`, for example: 
```
RPKI: parser encountered an invalid IP network: 155.186.0.0/16
RPKI: parser encountered an invalid IP network: 63.75.240.0/20
```
irrespective of whether the error is the network or a bogon AS. 

While fixing this I decided to extract a generic logging function so we don't repeat the same ceremony everywhere.

d5f15e9164412e87d663e3872c829d5c2e70434a refactors logging calls into a `context.debug(self, msg)` function in the Context module. I thought about putting this in util or somewhere else but context seems like the best place for it since it's tied to `context.debug_log`. 

24fdd9a2b3d5c655e32b3aca54a82b7742b631c0 splits the bundled prefix and ASN logging into separate log messages.

### DIFF
--- a/kartograf/rpki/fetch.py
+++ b/kartograf/rpki/fetch.py
@@ -61,13 +61,11 @@ def fetch_rpki_db(context):
                                 capture_output=True,
                                 check=False)
 
-        if context.debug_log:
-            with open(context.debug_log, 'a') as logs:
-                logs.write("=== RPKI Download ===\n")
-                if result.stdout:
-                    logs.write(result.stdout.decode())
-                if result.stderr:
-                    logs.write(result.stderr.decode())
+        context.debug("\n\n=== RPKI Download ===")
+        if result.stdout:
+            context.debug(result.stdout.decode())
+        if result.stderr:
+            context.debug(result.stderr.decode())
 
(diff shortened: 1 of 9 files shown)


kartograf/rpki/fetch.py
```diff
@@ -61,13 +61,11 @@ def fetch_rpki_db(context):
...
                                 check=False)
 
-        if context.debug_log:
-            with open(context.debug_log, 'a') as logs:
-                logs.write("=== RPKI Download ===\n")
-                if result.stdout:
-                    logs.write(result.stdout.decode())
-                if result.stderr:
-                    logs.write(result.stderr.decode())
+        context.debug("=== RPKI Download ===")
```
fjahr: This should ideally have `\n\n` in the front like the other sections. Maybe make another helper for sections like `context.debug_section("RPKI Download")` which then takes care of the rest.

jurraca: thanks, fixed by adding newlines in front, instead of another helper.

